PR #25, opened 2022-11-24
Adding Some Tests and Ran `cargo fmt`

colin99d:
Hello,

I was looking to add some tests and get your feedback on where you wanted testing added most, but then I ran cargo fmt and and that reformatted a bunch of code. I could add some Github CI over the formatting if you like.

### DIFF
--- a/src/enums/app_protocol.rs
+++ b/src/enums/app_protocol.rs
@@ -174,7 +174,6 @@ mod tests {
         let result = from_port_to_application_protocol(500);
         assert_eq!(AppProtocol::Other, result);
     }
-    // Please let me know if you want tests on every one, happy to do it so
 
     #[test]
     fn app_protocol_display_ftp() {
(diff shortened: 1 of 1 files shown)


GyulyVGC:
Don't worry for the formatting. I will do `cargo fmt` on my local copy without merging yours.

Thank you very much for the tests instead! Even if the tests you wrote are pretty trivial, it's still nice having them.

I would like to be able to test more the sniffing process, but this could be hard.

Another useful thing to do is to update the [workflow](https://github.com/GyulyVGC/sniffnet/blob/main/.github/workflows/rust.yml) including on Windows and Linux and their respective dependencies.

colin99d:
Would you like me to remove the cargo fmt stuff real quick?

GyulyVGC:
> Would you like me to remove the cargo fmt stuff real quick?

That'd be awesome, thanks!

colin99d:
Done!

PR #27, opened 2022-11-25
Adding CI for Linux and linters

colin99d:
This PR adds Linux tests and linters (Clippy and Rustfmt). It also looks into adding Windows CI, but currently there appears to be a blocking issue.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,20 +11,20 @@ env:
 
 jobs:
   fmt:
-  name: Rustfmt
-  runs-on: ubuntu-latest
-  steps:
-    - uses: actions/checkout@v2
-    - uses: actions-rs/toolchain@v1
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
-    - run: rustup component add rustfmt
-    - uses: actions-rs/cargo@v1
-      with:
-        command: fmt
-        args: --all -- --check
+    name: Rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - run: rustup component add rustfmt
+      - uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
 
   clippy:
     name: Clippy
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,13 +10,36 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build:
+  Test:
+    name: Mac
 
-    runs-on: macos-latest
+      runs-on: macos-latest
 
-    steps:
-    - uses: actions/checkout@v3
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: cargo test --verbose
+      steps:
+      - uses: actions/checkout@v3
+      - name: Build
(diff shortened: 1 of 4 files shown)


GyulyVGC:
I think this won't run because when running on Linux and Windows you need to specific which dependencies to install in the CI

colin99d:
Yup! I have it in draft to experiment!

colin99d:
@GyulyVGC it looks like if we want to use the Windows CI we will need to make a Npcap OEM account. See here:
https://github.com/rust-pcap/pcap/blob/main/.github/workflows/ci-steps.yml
https://npcap.com/oem/
I am not sure whether this is free or what is required to get one. For now should we ignore windows testing, and maybe get into contact with the npcap team?

GyulyVGC:
Mm, are you sure there is no workaround? 

Have you tried checking WinPcap as an alternative? You could check what [pcap](https://github.com/rust-pcap/pcap/tree/main/.github/workflows) do.

Thanks for your contributions!

colin99d:
https://github.com/nmap/npcap/issues/227

It looks like pcap had success with them in the past. I think for now we should remove the windows tests and put an issue on the npcap repo. If they are willing to work with us it would be much better to have the actual software we are using, rather than create a CI nightmare in the future by relying on incorrect software.

GyulyVGC:
Agreed!

GyulyVGC:
May I ask how the fmt and clippy steps work? If the workflow will fail also if I have introduced a not formatted line in my code I wouldn't like it that much.

colin99d:
Yeah they work as you described. I think they are valuable to have because it forces everyone to format the code before submitting a PR. Since rust makes this super easy `cargo fmt`, and since you can still merge a PR even if an action fails, I don't see the harm. I am working on having clippy ignore the `TCP` => `Tcp` thing right now.

GyulyVGC:
Sounds fair. We can keep them then.

colin99d:
Cool! Let me fix the clippy ignore and remove windows and we should be good. I suggest you leave an issue on NpCap: https://github.com/nmap/npcap, since I don't want to speak on behalf of your project. Also you got a discord or something so we can chat?

colin99d:
Everything on my end is done now!

GyulyVGC:
I still see clippy step failing. If I recall correctly the only warnings I had with clippy were related to camel case things.

GyulyVGC:
Whoops, just noticed the cause, are you able to exclude that check as well?

colin99d:
Yup!

colin99d:
Done!

GyulyVGC:
> [nmap/npcap#227](https://github.com/nmap/npcap/issues/227)
> 
> It looks like pcap had success with them in the past. I think for now we should remove the windows tests and put an issue on the npcap repo. If they are willing to work with us it would be much better to have the actual software we are using, rather than create a CI nightmare in the future by relying on incorrect software.

@colin99d this is just to inform you that today, after a bit of delays on their side, I've finally managed to setup a Windows workflow based on Npcap (#53).
Thanks for the idea!

colin99d:
> > [nmap/npcap#227](https://github.com/nmap/npcap/issues/227)
> > It looks like pcap had success with them in the past. I think for now we should remove the windows tests and put an issue on the npcap repo. If they are willing to work with us it would be much better to have the actual software we are using, rather than create a CI nightmare in the future by relying on incorrect software.
> 
> @colin99d this is just to inform you that today, after a bit of delays on their side, I've finally managed to setup a Windows workflow based on Npcap (#53). Thanks for the idea!

Lets go!!!!! Nice work.

PR #37, opened 2022-11-27
Instructions for NetBSD

0323pin:
`sniffnet` package merged into pkgsrc main branch, http://mail-index.netbsd.org/pkgsrc-changes/2022/11/27/msg264198.html


0323pin:
@GyulyVGC is it just the `Cargo.lock` you don't want or, is it also that you don't like to have a package for NetBSD?

GyulyVGC:
I think having it published on crates and having installers available for the moment in enough.

I have to get better knowledge about NetBSD and then I could reopen and merge this. Thank you anyway!

0323pin:
@GyulyVGC That's fine with me.

The thing is that NetBSD uses a different standard for paths. Pulling it from crates.io or, even building it locally with `cargo`  will yield binaries that won't run.

Basically, you need to tell `rustc` where to find system dependencies. Fortunately, it's enough to use `RUSTFLAGS` for one of the dependencies, `rustc` will then find evrything else.

Anyway, the package is merged and I'll keep it up-to-date.

GyulyVGC:
Thank you very much! I appreciate your work!

GyulyVGC:
Hey, I'm considering to add the NetBSD installation method in the README (for now it'd be on the secondary branch but I'll merge it into main at the next release).

Have you guys something like a badge for download counts?
In this way, it would be conformant to the layout of other installation methods I'm adding:

Screenshot 2023-01-11 at 16 01 00

I searched for a NetBSD badge on shields.io but found nothing.

Thanks in advance!

0323pin:
> I'm considering to add the NetBSD installation method in the README

That would be nice, thanks.

> Have you guys something like a badge for downloads count?

Not that I'm aware of  we don't track user downloads. Also, a lot of users on NetBSD prefer to build everything from source using the build infrastructure, very much like I'd guess Gentoo users do.

Another thing, pkgsrc supports, not only NetBSD but also, many other platforms, like Linux, MacOS and Illumos just to name a few.

Sorry.

0323pin:
Hi @GyulyVGC 
I've just updated the package on NetBSD, http://mail-index.netbsd.org/pkgsrc-changes/2023/02/08/msg269079.html

Would it be possible to have the `rodio`/audio dependencies behind a **feature** gate and make it optional?
`alsa` in not natively supported on NetBSD and it adds external dependencies to the package, including plugins to make it work with the native sound infrastructure.

`sniffnet`-1.1.0 works fine on my system but, crashes when I try to adjust the sound. It possible, I didn't configure the sound card correctly. Still, I think it would be nice to be able to have the audio part as an optional feature.

Thought?!

GyulyVGC:
I'm not aware of cargo features for executable crates that are not libraries.
How could a user enable/disable the feature? I think it should be done before the installation if I got what you mean.
In this case I should probably build two different binaries. Since I'm not very expert with those kind of feature I will. better investigate. In the meantime, your suggestions are welcomed.

GyulyVGC:
And of course an issue on this subject would be fine for me, if you want to open it. 
In this way other people with the same problem could share their point of view.

0323pin:
Yes, the idea was to have the notification sounds disabled by using `cargo build --no-default-features` at build time.

0323pin:
Before I go ahead and open an issue, this is the error I'm getting,

```
ALSA lib confmisc.c:855:(parse_card) cannot find card '0'
ALSA lib conf.c:5188:(_snd_config_evaluate) function snd_func_card_inum returned error: No such file or directory
ALSA lib confmisc.c:422:(snd_func_concat) error evaluating strings
ALSA lib conf.c:5188:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1334:(snd_func_refer) error evaluating name
ALSA lib conf.c:5188:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5711:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2664:(snd_pcm_open_noupdate) Unknown PCM default
thread 'thread_play_sound' panicked at 'called `Result::unwrap()` on an `Err` value: DefaultStreamConfigError(DeviceNotAvailable)', src/enums/[sound.rs:60](http://sound.rs:60/):72
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

GyulyVGC:
The problem is exactly that you miss that dependency for alsa-sys.
It's libasound-dev in Debian-like distro and libasound2-devel in Fedora.
What's the problem of adding an external dependency to the package?

0323pin:
@GyulyVGC 

> The problem is exactly that you miss that dependency for alsa-sys

Not exactly,

```
~ > cd /usr/pkg/lib
/usr/pkg/lib > ls | grep libasound
.rw-r--r-- root wheel  2.0 MB Mon Nov 21 05:39:08 2022 libasound.a
.rwxr-xr-x root wheel  971 B  Mon Nov 21 05:39:08 2022 libasound.la
lrwxr-xr-x root wheel   18 B  Mon Nov 21 05:39:08 2022 libasound.so ⇒ libasound.so.2.0.0
lrwxr-xr-x root wheel   18 B  Mon Nov 21 05:39:08 2022 libasound.so.2 ⇒ libasound.so.2.0.0
.rwxr-xr-x root wheel  1.2 MB Mon Nov 21 05:39:07 2022 libasound.so.2.0.0
```

Why they are not found, that's the puzzling bit.

> What's the problem of adding an external dependency to the package?

That's not a problem, it's just that NetBSD doesn't use `alsa`.

GyulyVGC:
Mm, that's really weird.
Having in mind your problem, today I tried Sniffnet on a fresh Fedora VM to see if the problem arises on other platforms and it seems to work even on RPM-like distro.
I don't know honestly if it'd be meaningful to introduce a feature to make the app sound-less, considering that on all the other platforms I tried it seems to work well.

0323pin:
I'll try to dig a bit more into it. I've already told `rustc` where to find the libraries with `RUSTFLAGS+=    -C link-arg=${COMPILER_RPATH_FLAG}${BUILDLINK_PREFIX.alsa-lib}/lib`

Without this, it complains about missing runtime dependency `libasound.so`. So it seems the compiler is happy but, the resulting binary is not.

> I don't know honestly if it'd be meaningful to introduce a feature to make the app sound-less

This is up to you, I'm seriously biased since I dislike all sorts of notifications and switch them all off on all my devices.

GyulyVGC:
It'd be awesome if you could dig into it and let me know if you have any news, thanks a lot.

For which concerns a sound-less experience, the app already supports this functionality: even with notifications set you can disable their sound by setting the volume to zero (which may cause a crash in your specific case) but also by setting the sound to "none" for each notification kind (which will not crash your app).
So, the problem is basically just the compatibility, not the the usability.

0323pin:
> So, the problem is basically just the compatibility, not the the usability.

Correct, the application works as it should as long as one doesn't try to adjust the sound.

I've asked internally, http://mail-index.netbsd.org/tech-pkg/2023/02/10/msg027293.html if someone has a clue or suggestion.

GyulyVGC:
[Done](https://github.com/GyulyVGC/sniffnet/commit/09d60d2a2d4e29c2d3ba5c8836eb61e750677f55).

Sorry for the delays and the first rejection, but I'm starting to appreciate the possibility to have sniffnet available for more systems along the way.

Thanks for maintaining the package of NetBSD 👍

GyulyVGC:
@all-contributors please add @0323pin for platform.

GyulyVGC:
Also added to [sniffnet.net/download](https://www.sniffnet.net/download/)

GyulyVGC:
I just noticed you also mentioned FreeBSD on Reddit... which is the command to install from there?

0323pin:
@GyulyVGC

> So... the sound problem hasn't be solved, right?

Honestly, I don't know. Haven't tried sound since. 1.2.1 is building right now and I can try.
If you're interested `pulseaudio` works fine on NetBSD, so that's a possibility. Although, I don't use pulse myself and prefer to avoid it. Another alternative (preferable) would be `portaudio`.

Also, I've added NetBSD support to `cpal` [back in October 2022](https://github.com/RustAudio/cpal/pull/693), so in principal `rodio` should work.

> However, I'm going to include NetBSD in the docs anyway.

Thanks! I think it's nice to show cross-platform support :)

0323pin:
> I just noticed you also mentioned FreeBSD on Reddit... which is the command to install from there?

I don't use FreeBSD but, saw on Repology that there was a package for `sniffnet`. 

But according to the docs, https://docs.freebsd.org/en/books/handbook/ports/
should be `pkg install sniffnet`

GyulyVGC:
Could you tag the maintainer of the FreeBSD package if you know him/her?
So that I can also include him/her in the contributors and ask for info when needed.

0323pin:
You can find it in [Repology](https://repology.org/project/sniffnet/versions) and from the [commit history](https://github.com/freebsd/freebsd-ports/commits/main/net/sniffnet) doesn't look like the maintainer uses GitHub.

Sorry for not being of better help, I've never used FreeBSD and honestly, I have no interest in it.

OT: Building 1.2.1 spits-out a warning you may or, may not be aware of,

```
warning: unused variable: `report_path`
   --> src/gui/types/sniffer.rs:272:17
    |
272 |             let report_path = get_report_path();
    |                 ^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_report_path`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `url`
   --> src/gui/types/sniffer.rs:293:13
    |
293 |         let url = web_page.get_url();
    |             ^^^ help: if this is intentional, prefix it with an underscore: `_url`
```

I'm using Rust-1.70.0

GyulyVGC:
Oh shoot, that happen only on OS different from Windows, macOS, Linux, due to conditional compilation: https://github.com/GyulyVGC/sniffnet/blob/6fa13cdcd34d5bbbce4ea2bb5abdef0af68f3969/src/gui/types/sniffer.rs#L272

GyulyVGC:
It should not involve important functionalities though, so don't worry

0323pin:
You could replace `#[cfg(target_os = "linux")]` with `#[cfg(any(target_os = "linux", target_os = "netbsd"))]`, we do support `xdg-open` ;)

0323pin:
@GyulyVGC

F**k, excuse my language :)
Sound it's working, just tested. The issue is something completely different.

So, to set-up the desired interface I have to use `sudo` (actually, I don't use `sudo`, I use [`please`](https://gitlab.com/edneville/please) but, that doesn't matter). The thing is that `.asoundrc` is owned by the user and not root. I can adjust the sound if I re-launch the app as user after picking the network interface. So, all good with sound :)

GyulyVGC:
Holy moly, awesome! 👍

0323pin:
I'm assigning myself as package maintainer for this one now, so I get RSS notifications when new versions are released.

Nice talking to you :)

GyulyVGC:
Same for me!

PS: I will solve the conditional compilation ASAP, but no worries since it only involves opening links of the footer and opening the output report file (these should be the only things that currently don't work on NetBSD)

0323pin:
The update is merged into the main branch now.

If you don't mind, I'd like to leave a feature request. I compile all my packages from source and for `sniffnet`, I have a personal patch to reduce the font size as I find the default too big. Would it possible to make it adjustable?

GyulyVGC:
Iced, the library I use for the GUI, is undergoing major changes for what concerns text representation and is also switching to a new graphical renderer.

So, I think it's not worth it to lose time on something that is changing and will probably be solved by the upcoming releases of iced.

0323pin:
That's fine, I agree with that. If `iced` is undergoing major changes, let's wait until they are done. No rush, I'll keep my patch 👍

GyulyVGC:
Thanks for understanding!

GyulyVGC:
The warnings and related problems should have been fixed in f47bfaa15de60794397e09ef3279be36592f6d7f

0323pin:
Thank you!

GyulyVGC:
> So, to set-up the desired interface I have to use `sudo` (actually, I don't use `sudo`, I use [`please`](https://gitlab.com/edneville/please) but, that doesn't matter). The thing is that `.asoundrc` is owned by the user and not root. I can adjust the sound if I re-launch the app as user after picking the network interface. So, all good with sound :)

I forgot to mention a thing... I don't know if it's possible on NetBSD but on Linux it's enough to set the correct capabilities to the binary (avoiding to launch the app in sudo mode): 

```sh
sudo setcap cap_net_raw,cap_net_admin=eip <your/Sniffnet/executable/path>
```

This is included as post installation script on Linux and maybe you could do the same thing with your package?

Issue #60, opened 2023-02-07
🌍 Help me translating Sniffnet in your native language 🌍


GyulyVGC:
In v1.1.0 I introduced the possibility to select a language for the GUI.
At the moment, only English and Italian are supported, but my goal is to make Sniffnet available in more languages.
If you are a native speaker of one of the languages below, it'd be awesome if you could help me translating the application in your language.

The phrases to translate are just about 60 and are pretty short, so it shouldn't take too much time.
All you have to do is to add the translations to [this](https://github.com/GyulyVGC/sniffnet/blob/main/src/utility/translations.rs) file. 

<div align="center">

Language | Two-letters code | Flag | Status
-|-|-|-
English | EN | 🇬🇧 | ✅
Italian | IT | 🇮🇹 | ✅
German | DE | 🇩🇪 |
French | FR | 🇫🇷 |
Spanish | ES | 🇪🇦 |
Portoguese | PT | 🇵🇹 |
Chinese | ZH | 🇨🇳 |
Japanese | JA | 🇯🇵 |
Russian | RU | 🇷🇺 |

</div>

I included in the table the most common languages of my users, but if you speak a different language I will be happy to include your translations as well!

Thanks in advance to everyone!

acolombier:
I have started adding french as part of the PR #64, however, as mentioned in there:

> Worth to mention that although native speaker, it has been a few years I have written my own language in a formal way, so if there is any fellow french speakers that could help review it, that would be massively appreciated!

Charpy:
> I have started adding french as part of the PR #64, however, as mentioned in there:
> 
> > Worth to mention that although native speaker, it has been a few years I have written my own language in a formal way, so if there is any fellow french speakers that could help review it, that would be massively appreciated!

Hi ! as I'm a french native I did some review on the french translation. I'm not very used to Github features so I hope everything worked well with "branches, pull up, commit, merge"... I kept the reviewed file locally so feel free to contact me if needed.

AmadeusGraves:
Added Spanish. Cheers!

Kiruyuto:
I've added polish translation at #78, but yoy already answered that PR 😅

Also a question regarding other translations - Since I don't speak these languages - is the space before " : " intended? @AmadeusGraves 
eg.
https://github.com/GyulyVGC/sniffnet/blob/e918c062f39599950e95fb88bfd66c4f46689108/src/utility/translations.rs#L166
[...]

PR #64, opened 2023-02-08
Adding french translation

acolombier:
Answering to the call in the issue #60, this PR adds translation for french.

Worth to mention that although native speaker, it has been a few years I have written my own language in a formal way, so if there is any fellow french speakers that could help review it, that would be massively appreciated!

### DIFF
--- a/src/enums/language.rs
+++ b/src/enums/language.rs
@@ -7,6 +7,8 @@ pub enum Language {
     EN,
     /// Italian.
     IT,
+    /// French.
+    FR,
 }
 
 impl Default for Language {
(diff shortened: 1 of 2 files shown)


GyulyVGC:
That's amazing! 
If you know other French people you could ask directly to them to give it a look. However, if you speak French don't worry too much about the level of formalism, your translation will be surely good enough!

GyulyVGC (commented):
I know nothing about French, but it seems to me that the last phrase refers to "minutes" instead of "notifications"

acolombier:
You're correct - thanks for catching that! Should be updated now.

GyulyVGC:
Thanks a lot for the translation, now all checks pass and everything seems good.
I will do the last graphical checks on the GUI and then the PR will be merged.

Katew212 (commented):
[Y]()

Issue #62, opened 2023-02-08
Add AS Name into Results


m-cameron:
Please consider to append ASN information into the results.

You can get the free ASN database from https://lite.ip2location.com/database-asn?lang=en_US

GyulyVGC:
This is actually one of the features I will implement in v1.2.0.
Other info that will be added are:
- MAC addresses
- rDNS lookups of the remote address

I think I will implement a more detailed page for each of the sniffed connections with these info and their ASN.
If you have other ideas, you can share them without any problem.

acolombier:
I am very interested in the rDNS feature! Is it something you need help to develop? I'd be happy to give it a go

GyulyVGC:
> I am very interested in the rDNS feature! Is it something you need help to develop? I'd be happy to give it a go

Thank you very much for your proposal, but I've already tried to implement this feature and it is pretty straightforward thanks to a crate that performs rDNS lookups.

For v1.2.0 I have in mind a larger set of features which will include a new page with a search bar to select a connection to inspect and it's something pretty big and I would like to start it myself.

I will tag you when I'll have something ready (probably in some weeks or a few months) and at that point it'd be amazing to collaborate.

In the meantime, if you really want to help, it'd be nice to write some unit tests for the app (see [here](https://github.com/GyulyVGC/sniffnet/blob/main/src/enums/app_protocol.rs#L160) for an example of unit tests by @colin99d, to be written at the end of the file with the functions interested by tests).

acolombier:
That sounds good to me - I will submit some for the `enums` crate - can see few low hanging fruits in there!

GyulyVGC:
> That sounds good to me - I will submit some for the `enums` crate - can see few low hanging fruits in there!

Yep, that should be an easy way to start, thanks very much!

GyulyVGC:
[A new release](https://github.com/GyulyVGC/sniffnet/releases/tag/v1.2.0) including Autonomous Systems and rDNS lookups has just been published!

PR #65, opened 2023-02-08
chore: add some unit test on notification

acolombier:
As suggested in the issue #62 , I tasked myself with increasing test coverage and thought I will get started with notifications.

### DIFF
--- a/src/structs/notifications.rs
+++ b/src/structs/notifications.rs
@@ -1,4 +1,4 @@
-use crate::enums::sound::Sound;
+use crate::enums::{byte_multiple::from_char_to_multiple, sound::Sound};
 use crate::ByteMultiple;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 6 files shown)


acolombier:
Since this task requires some minimal refactoring in order to split the code in smaller independent sections, I thought I will start with `byte threshold` first and wait for your feedback before going ahead and doing all of them.

GyulyVGC:
Sounds good. Just leave me the refactor tasks. The ideal thing is that you exclusively work inside "tests" modules.

GyulyVGC:
However, if you already started refactoring keep going. 
Just let me know when and where you need it next time.

GyulyVGC:
At a first read it seems a very good start btw! 
It's nice you decided to put that function in the notifications struct, feel free to do the same for the other notifications.
Thanks!

acolombier:
Do you mind pointing out which bits your are referring to? Do you mean the `Default` trait implementation?

src/structs/notifications.rs
```diff
@@ -44,6 +32,34 @@ pub struct PacketsNotification {
...
+            threshold: None,
+            sound: Sound::Gulp,
+            previous_threshold: 750,
+        }
+    }
+}
+
+impl PacketsNotification {
+    /// Arbitrary string constructor. Will fallback values to existing notification if set, default() otherwise
+    pub fn from(value: &str, existing: Option<Self>) -> Self {
```
GyulyVGC, line 47: This function content should be the same of the respective previous segment of code in settings.rs (e.g, when text field is empty, consider it as a zero and not as a None; I'd like this functions to have the same content and behaviour as the previous version of the code)

acolombier:
I have reverted the `PacketsNotification` and `BytesNotification` behaviors. I kept the `trim` addition so there is more edge cases covered, although the UI doesn't let you input anything else.

GyulyVGC:
Thanks! Once also the formatting will be fixed I'll review it! 👍

GyulyVGC:
I just fixed the formatting (note that this can be automatically done with `cargo fmt`) and reviewed the PR and you did a great job.
It's nice you decided to start from the text input fields which are the most vulnerable part of the app.
Tyvm!

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @acolombier for test translation

PR #78, opened 2023-02-18
Add Polish translation

Kiruyuto:
Added Polish 🇵🇱 translation, should be about 99.99% correct 👍

### DIFF
--- a/src/enums/language.rs
+++ b/src/enums/language.rs
@@ -11,6 +11,8 @@ pub enum Language {
     FR,
     /// Spanish.
     ES,
+    /// Polish.
+    PL,
 }
 
 impl Default for Language {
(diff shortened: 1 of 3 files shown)


GyulyVGC:
Wooo! Thank you very much! 💯 

I will add some lines to your code to make it compile and I will graphically check that everything is ok, since I see that Polish uses some non-standard character and the library I use for the GUI currently supports special characters only if included in the font specified during development.

Kiruyuto:
There isn't many fonts that support these characters by default so I wouldnt be surprised.
Let me know if I should change it into some standard characters, instead polish-specific

GyulyVGC:
Everything seems fine to me.
[This](https://www.1001fonts.com/inconsolata-font.html) is the font I used and it seems to have all the needed characters.

Kiruyuto:
Yup, I believe everything should be fine!

image

GyulyVGC:
@all-contributors please add @Kiruyuto for translation.

PR #94, opened 2023-03-03
Add Ukrainian translation

captain-yossarian:
Add Ukrainian translation


src/utility/translations.rs
```diff
@@ -584,6 +638,7 @@ pub fn deep_sea_translation(language: Language) -> &'static str {
         Language::ES => "Para sumergirse en el tráfico de la red",
         Language::PL => "Aby zanurzyć się w ruchu sieciowym",
         Language::DE => "Um in den Netzwerkverkehr einzutauchen",
+        Language::UA => "Оригінальний, темний мотив Sniffnet",
```
GyulyVGC, line 641: I was rapidly looking at the translations and found out that this one I equal to one of the preceding and I think it's not in the correct spot

GyulyVGC (changes requested):

src/utility/translations.rs
captain-yossarian: Good catch, will fix

captain-yossarian: @GyulyVGC  done

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @captain-yossarian for translation.

PR #98, opened 2023-03-05
Started adding Dutch translation. Searching for any other fellow Dutchmen to continue 🇳🇱

keiveulbugs:
As title says, I started adding Dutch to it. I completed it till line 333, but got busy and currently don't have time to complete it further. Hopefully soon I do, but any other fellow Dutchmen feel free to pick it up. Might be good to read the current translations thus far as sometimes I used a better sounding translation instead of a direct translation (i.e. Netwerk activiteit (network activity) instead of netwerk verkeer (Networktraffic).


GyulyVGC:
Thanks!

PR #113, opened 2023-03-23
Added Romanian language and translations

ervinpopescu:
There is a little problem, the font you are using doesn't support diacritics(ă,î,â,ș,ț) which are used in Romanian.


GyulyVGC:
Don't worry about that: I manually subsetted the font, so I will include all the used characters later!
Thank you very much for the translation!

GyulyVGC (approved):

ervinpopescu:
This looks great, thanks for the corrections!

PR #134, opened 2023-04-12
Add portuguese translation

franciscoBSalgueiro:


GyulyVGC:
A PR for Brazilian Portuguese has also been opened (#136).
Is it meaningful in your opinion to have both languages featured or you guys can agree on a single language? Do they differ substantially?

franciscoBSalgueiro:
It's the same language, so any portuguese speaking person would be able to read both. That being said, you can definitely tell one was written by a Brazilian and another by a Portuguese. I've seen some apps have both a pt_PT and pt_BR translation, but if you only want to have one, I think any of them would work.

GyulyVGC:
If standard Portuguese is also comprehensible by Brazilians, I'd prefer to keep only one.
I kindly ask to @vrmiguel to review this PT translation and let me know if some term is not clear. 
If something is not clear, may you find a version that suits both?
Thanks in advance guys!

GyulyVGC:
Since #136 has been closed, I'll proceed merging this PR!

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @franciscoBSalgueiro for translation

PR #136, opened 2023-04-13
i18n: add translation for Brazilian Portuguese

vrmiguel:
Adds a translation for Brazilian Portuguese

Edit: oh well, just noticed #134 is a translation for European Portuguese. I'll rename the enum variant in this PR to PT_BR and then I assume both will be able to be merged


GyulyVGC:
Is it meaningful in your opinion to have two different translations? Do the two languages differ substantially?

Issue #137, opened 2023-04-13
segfault on macos


blinsay:
**Describe the bug**

I just learned about sniffnet through the accelerator program (congrats!) and tried installing it on macos through homebrew and Cargo. I get a segfault after the welcome banner. Happy to help diagnose more if you let me know what is useful to look at.

**To Reproduce**

```
$ sniffnet

  /---------------------------------------------------------\
 |     _____           _    __    __                  _      |
 |    / ____|         (_)  / _|  / _|                | |     |
 |   | (___    _ __    _  | |_  | |_   _ __     ___  | |_    |
 |    \___ \  | '_ \  | | |  _| |  _| | '_ \   / _ \ | __|   |
 |    ____) | | | | | | | | |   | |   | | | | |  __/ | |_    |
 |   |_____/  |_| |_| |_| |_|   |_|   |_| |_|  \___|  \__|   |
 |                                                           |
 |                   ___________                             |
 |                  /___________\                            |
 |                 | ___________ |                           |
 |                 | |         | |                           |
 |                 | | v1.1.2  | |                           |
 |                 | |_________| |________________________   |
 |                 \_____________/   by Giuliano Bellini  )  |
 |                 / ''''''''''' \                       /   |
 |                / ::::::::::::: \                  =D-'    |
 |               (_________________)                         |
  \_________________________________________________________/
[1]    13150 segmentation fault  sniffnet
```
```
$ brew info sniffnet
==> sniffnet: stable 1.1.2 (bottled), HEAD
Cross-platform application to monitor your network traffic
https://github.com/GyulyVGC/sniffnet
/opt/homebrew/Cellar/sniffnet/1.1.2 (9 files, 11.7MB) *
  Poured from bottle using the formulae.brew.sh API on 2023-04-13 at 00:52:28
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/sniffnet.rb
License: Apache-2.0 or MIT
```

**Desktop (please complete the following information):**

```
$ sw_vers
ProductName:		macOS
ProductVersion:		13.3.1
BuildVersion:		22E261
```

GyulyVGC:
I think this is a duplicate of #109.
It's weird tho because in that case only Homebrew installation was affected.
However, as I mentioned in #109, the problem will hopefully be solved with the next release of iced.

GyulyVGC:
Are you sure you launched sniffnet also from the cargo binary?

blinsay:
[...]

PR #143, opened 2023-04-13
Update iced to 0.9.0 which also fixed the crash on macos

shenjiangqiu:
Update the dependence: iced from 0.8 to 0.9, which also solved those issues: #137 , #109 ,


GyulyVGC:
Thanks very much for the effort but I already updated it on the branch iced_v0.9 which will we merged soon!

PR #151, opened 2023-04-15
feat: add Russian translation

amarao:
N.B. It's broken and I can't understand why. Letters 'Р' and 'ы' are not displayed. Please help, if you can.

image


GyulyVGC:
> N.B. It's broken and I can't understand why. Letters 'Р' and 'ы' are not displayed. Please help, if you can.

Don't worry, I'll add the missing characters to the font file.
Thank you very much for the translation 🙏

GyulyVGC:
Please, if you can let me know if everything looks good now

amarao:
Yes, it looks nice! Even `ё` (which was also missing) is present. Thank you!

amarao:
Btw, I feel that manual letter update is a bit ... too manual? May be it can just grab all letters from the translation source file (May be even not at Rust level, just list of unique characters in the text file with .rs extension).

GyulyVGC:
> Btw, I feel that manual letter update is a bit ... too manual? May be it can just grab all letters from the translation source file (May be even not at Rust level, just list of unique characters in the text file with .rs extension).

Yep, it's exactly what I was meaning with manual. 
I grep all the characters of the src folder and add them from the full font into the subset font with a command line utility.

GyulyVGC:
@all-contributors please add @amarao for translation

PR #158, opened 2023-04-19
Added Persian translation

4r7if3x:
This PR addresses a few improvements in the font and language support:

**Change Log:**
- Resolved the character rendering issue by incorporating font support for an extensive selection of languages
- Added Persian as the first RTL language along with its respective translations
- Reorganized the languages and their translations according to their ISO codes to improve readability and consistency

**Known Issues:**
- RTL strings do not render correctly ( related to iced-rs/iced#33 )
- The width of radio labels in the Language settings needs adjustment


GyulyVGC:
Thanks for your contribution.

Just a few notes:

- before adding new font files, have you tried to use the fonts in the font/full folder?
- I cannot include all those fonts file, as the size of the binary would be too high

4r7if3x:
> * before adding new font files, have you tried to use the fonts in the font/full folder?

Yes, those won't work for Persian as well as some other languages. Besides, there is no single font for everything. Having them merged also won't help with the volume, nor it's a straightforward process.

> * I cannot include all those fonts file, as the size of the binary would be too high

I believe we can also load them on-demand from CDN at the run-time. Do you have any other suggestions?

4r7if3x:
On second thought, what if we don't use a custom font at all, but rely on a widely-used, pre-installed default font on the client side? Would _Iced_ support system fonts just by specifying their names?

GyulyVGC:
I don't think so... until iced will not support font fallback I think there is little we can do.
I'd suggest waiting for iced to support it before merging this PR.

4r7if3x:
I addressed the issue with the executable size and kept the original font as a fallback.

In near future, I tend to assist with packaging the app for more platforms and installer tools. At that point, we may opt to install the custom fonts during setup, considering the approaches taken by traditional desktop apps, but I don't like the idea of installing other language fonts on the client's system.

I still like the idea of relying on pre-installed fonts if we can manage to display the language names on the settings page whether the language font exists or not, perhaps by bundling a limited-character custom font. However, this method requires a fallback solution, such as the dynamic loading that I just introduced in this PR, so it can handle unexpected selections of curious minds.

GyulyVGC:
I'll give a more in-depth look to this PR in the next few days.
It seems a good start, but I'm still convinced that there could be a better solution without involving a CDN.
I could try keeping the original font file for the other languages while using for Persian a subsetted version of the font file you submitted yesterday.
I know it's not a very scalable solution if in the future new arabic translations will be submitted, but it's the best thing that comes to my mind.

Also, it'd be very appreciated if you could help with packaging.
In particular, I'd like to come up with a script to automate the packages build for the different platforms, and I'd prefer to generate not just a binary but a bundle including a desktop icon.

4r7if3x:
> I'm still convinced that there could be a better solution without involving a CDN.

I understand, yet traditionally custom fonts were either bundled in the executable or in the resource files that come along with it, if the app didn't use system fonts.

I understand... But traditionally, when the application didn't adopt system fonts, custom fonts have either been bundled within the executable or included in the accompanying resource files. Loading fonts from CDN however, is a common practice in hybrid applications.

Nonetheless, we can still merge glyphs from other language fonts into the subset font that you have already created, since only a few additional characters would be introduced, although we may encounter some technical difficulties in the process.

> I'd prefer to generate not just a binary but a bundle including a desktop icon.

Sure, that was actually my main objective. I can certainly help with the other aspects as well.

GyulyVGC:
Sorry for the delays, but I'm currently busy implementing new features for the next version.

Tomorrow I'll merge the font file with Arabic characters in the full font file and I'll subset it keeping only the used glyphs.
My plan is to keep only the modifications you made with [this commit](https://github.com/GyulyVGC/sniffnet/commit/24692ef31c82c4eb4fd6d1adf84c112c8798f4c5).

GyulyVGC:
I've added the arabic glyphs to the original font files.
Let me know if everything seems ok to you.
If it's all good, I'll proceed merging the PR.

GyulyVGC:
@all-contributors please add @4r7if3x for translation.

GyulyVGC:
@4r7if3x are you still willing to help packaging the app? 
I'd especially need some kind of automation to package deb and rpm packages for Linux.
If you are in, I can assign you an issue giving more details.

4r7if3x:
> @4r7if3x are you still willing to help packaging the app? I'd especially need some kind of automation to package deb and rpm packages for Linux. If you are in, I can assign you an issue giving more details.

Sure! Please keep in touch via email and let me know about the details over there.

PR #160, opened 2023-04-21
Added Greek translation

abousis:
Added Greek translation, although Greek letters do not appear at all. I suppose it's the font.


GyulyVGC:
Yep, it's a problem of the font. 
I'll check if the full font file contains greek characters and I'll let you know.
Thanks!

GyulyVGC:
I fixed the font problem but I noticed that in some points some greek sentences seems too long and they go on a new line.
For the first picture, could you try using a shorter sentence to convey the same meaning?
For the second picture, I'll try to do something myself.
 (It's weird what's happening because in the first case, for example, the Turkish sentence is longer but it displays without problem in the same amount of space)

Screenshot 2023-04-21 at 18 42 51
Screenshot 2023-04-21 at 18 42 42

abousis:
Thank you for the swift response. Checking, and I will revert with an update.

abousis:
I've rewritten the sentence in various ways, it seemed to me character "Θ" was causing this, but it was strange in any case. Also there are longer sentences in Greek, so I suppose it has to do with the font and this specific character.
It seems it's displayed properly now.

screenshot_20230421_212226

screenshot_20230421_212239

GyulyVGC:
Thanks.
I'm wondering why you are seeing those black rectangles instead of the palettes pictures...
Which system are you using? Can you correctly see the country flags?

abousis:
This is on the main branch, if I build the application using the glow-renderer branch as you suggested on README.md, then everything is displayed properly.

screenshot_20230421_230043

src/translations/translations.rs
```diff
@@ -1212,7 +1212,7 @@ pub fn threshold_translation(language: Language) -> String {
         Language::TR => "Eşik: ".to_string(),
         Language::RU => "Порог: ".to_string(),
         Language::PT => "Limite: ".to_string(),
-        Language::EL => "Όριο: ".to_string(),
```
GyulyVGC: I've replaced this character with its lowercase version and the problem of the second picture I sent you before is now solved

GyulyVGC:
If you can confirm everything is now ok, I'll proceed merging the PR

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @abousis for translation.

PR #180, opened 2023-05-15
Automated build and packaging

4r7if3x:
Extensive testing on different platforms is required.


GyulyVGC:
Here is a summary of the problems / TODOs:
- I temporarily removed all the targets / build targets since they were causing problems I don't know how to solve
- I've included the Debian-based distro dependencies and those dependencies should also be added to the deb package (the same should also applies to the rpm package)
- The installation of cargo-generate-rpm seems to have some problems:
image
- I've commented out the Windows workflow since it need a secret to install npcap dll, but secrets are not passed to PRs from a fork (I'll personally solve this one as it'll be merged)
- I've commented out the macOS workflow since it needs to generate a wrapper to launch the bundle in sudo mode (also for this point, I'll personally think about it later)

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @4r7if3x for platform.

PR #191, opened 2023-05-18
Additional German translations added

luca3s:
After seeing the discussion regarding the name "Administrative entity" i've made another commit where it translates to "Autonomous System Name". If you decide to stay with your original name you can take [this](https://github.com/GyulyVGC/sniffnet/commit/5d63309bf0efafbae63b481a28a1248fa3a14b68) commit instead.


GyulyVGC:
Thanks!

GyulyVGC (approved):

PR #206, opened 2023-05-21
Correction of awkward Korean expressions

eatingdm:
'한국인' is a Korean expression that refers to a person living in Korea, and '한국어' is correct.


GyulyVGC:
Awesome, thanks!

GyulyVGC (approved):

PR #213, opened 2023-05-22
Swedish translation

vcrn:
This PR is to add Swedish as a language option, as requested in Issue #60.

There's one thing that needs to be fixed though, and one to be discussed:
* The letters `Ä`, `Å` and `å` can't be displayed, they only result in a white space. I've added these to `resources/fonts/full/subset_characters.txt`, but the issue persists, so I'm not sure what the cause is. Is there a certain order in which they need to be placed? Because now, I've put them in arbitrary locations in `subset_characters.txt`.
* I started out by creating a 4th column for Swedish in the language selection page, but it became a bit cramped with 4 columns, so I went with adding it to the 1st column. Does this, and the placement, seem alright with you?

I've also made (what seems to me to be) some small corrections of other texts.


GyulyVGC:
Hi @vcrn thank you very much.

For point 1): don't worry, I'll fix it personally. (It's also necessary to include those letters in the font file but I use a program to do it, so don't worry of doing it yourself).

For point 2): I'll fix it as well!

GyulyVGC:
Update:
- added missing characters to the font files
- restructured language selection page
- renamed Swedish language code to `SV` (`SE` is the code for Sweden, the country)

If you confirm everything looks good to you as well, I'll proceed merging

GyulyVGC (approved):

vcrn:
Great, I'll check it out today or tomorrow!

GyulyVGC:
The library I use for the GUI has a very basic text handling as of today so I don't think it supports a similar feature.

After having played a bit with the relative widths of the filter columns I have come up with sizes that correctly display all the columns title in one line without cuts, in all the currently available languages.
Thank you very much for having pointed out the problem (I noticed it happened for Korean as well).
Confirm everything's good when you can.

PS: you didn't forget the `Å` actually, I removed it by edited that file by grepping all the used characters in the src folder.

vcrn:
Great, now it looks good in full-screen on both laptop screens!

GyulyVGC:
@all-contributors please add @vcrn for translation.

vcrn:
> @all-contributors please add @vcrn for translation.

Thanks, appreciate it! 

I'll see if there's something else I can help out with. Right now I'm looking into what makes one of my laptops running Manjaro able to display the flag-icons using `wgpu`, while my laptop that runs Arch can only do it using `glow`. Will see if it's due to hardware or some missing dependency

GyulyVGC:
> Thanks, appreciate it!
> 
> I'll see if there's something else I can help out with. Right now I'm looking into what makes one of my laptops running Manjaro able to display the flag-icons using `wgpu`, while my laptop that runs Arch can only do it using `glow`. Will see if it's due to hardware or some missing dependency

I must be thanking you man!
I think that problem is "normal", Iced documented it a lot and suggests to use `glow` in case of problems.
However, they are introducing a new software renderer that should be "universal" (but someone already reported on their Discord that in some cases this bug happens as well)

vcrn:
> I must be thanking you man! I think that problem is "normal", Iced documented it a lot and suggests to use `glow` in case of problems. However, they are introducing a new software renderer that should be "universal" (but someone already reported on their Discord that in some cases this bug happens as well)

Cool, I'll check it out!

In case it's of any interest, my laptop that could display the icons while running Manjaro can still display them after a fresh install of Arch, so it seems like it's something on the hardware level that is making my other Arch laptop unable to display them.

GyulyVGC:
Yep, absolutely.
I forgot to mention that this problem usually has to do with the graphical card / adapter of the machine.

Issue #223, opened 2023-05-28
Typo on Windows 10 Start Menu


bgodole:
When installing Sniffnet under Windows 10, a folder and a shortcut to the software is created in the Start Menu. The shortcut reads "Sniffet" instead of Sniffnet.

![sniffnet](https://github.com/GyulyVGC/sniffnet/assets/15838230/4e8c9957-558a-440e-942c-0ee9d960abaf)

GyulyVGC:
Thank you very much for having pointed out!

PR #230, opened 2023-05-30
Update main.wxs to fix bug #223

q5sys:
Fix bug - https://github.com/GyulyVGC/sniffnet/issues/223


GyulyVGC:
Hi @q5sys, the app packaging strategy is undergoing a complete restructure and as a consequence also the `main.wxs` file.

Thank you very much for your intention to help anyway!

Issue #63, opened 2023-02-08
Avoid dealignment of report when longer addresses are received


GyulyVGC:
Currently, when addresses longer than 25 characters are received, the report row will look like this:

![Screenshot 2023-02-08 at 08 35 38](https://user-images.githubusercontent.com/100347457/217464235-cdbbbc48-eba7-4d48-9092-71bb952b2336.png)

causing a dealignment of its text fields.

I was not able to reproduce this case on my PC and I think it's not a so common case, but it was first documented in issue #48.

I partially solved that issue introducing a horizontal scrollbar, as you can see from the picture, but I'd like to find a better solution for the long term.

The best solution would be to use a table widget, but [iced](https://github.com/iced-rs/iced) doesn't support it yet.

PR #241, opened 2023-06-03
added Romanian translation v1.2

ervinpopescu:
New characters need to be added in the font @GyulyVGC :) One question, the table header doesn't need to be updated?
image


GyulyVGC:
Thank you very much for the update.
I'll add the missing characters.

For what concerns the table header, I don't translate it yet because of #63.
Translating the table header means having the headings of different lengths, which could possibly cause a dealignment.
I have to implement a different mechanism for the table (for the moment it's just made of aligned ASCII characters).

GyulyVGC (approved):

PR #257, opened 2023-06-07
Final adjustments

4r7if3x:


resources/packaging/linux/sniffnet.desktop
```diff
@@ -3,8 +3,8 @@ Version=1.0
 Type=Application
 Name=Sniffnet
 Comment=Application to comfortably monitor your network traffic
-Categories=Network;
+Categories=Network;Utility;
 Icon=sniffnet
-Exec=/usr/bin/sniffnet
```
GyulyVGC, line 8: what was the problem with this one?

4r7if3x: Nothing, but `/usr/bin` is already in system `PATH`...

P.S. I'll send a new PR, my previous commit missed a thing and this PR is again not updated automatically...

4r7if3x: Yes please & No, it'd be fine.

GyulyVGC: If someone install the app also via cargo or homebrew... which version will open just typing Sniffnet? If no problem was caused, I'd leave thing as they were

Issue #243, opened 2023-06-04
[Feature Request] Option to provide path to own geoip database.


IncSW:
**Is your feature request related to a problem? Please describe.**
To have better ip information accuracy, if you have paid database

**Describe the solution you'd like**
Options like `-geoip-country /path/to/country.mmdb`, `-geoip-asn /path/to/asn.mmdb`

GyulyVGC:
Thanks for the suggestion.
This is something I could consider for the next major release.

GyulyVGC:
Hey @IncSW, a fix to this issue has just been merged to the main branch (implemented as part of #365): now both ASN and Country database paths can be specified via in-app settings.

This and other changes will be officially released in version 1.3.

PR #272, opened 2023-06-12
feat: support --help and --version CLI arguments

### LINKED ISSUE #264: sniffnet doesn't respect `--help` and `--version` flags
**Describe the bug**
Running `sniffnet --help` or `sniffnet --version` should display information in the terminal without starting the GUI.

**To Reproduce**
1. Run `sniffnet --help` or `sniffnet --version`

**Expected behavior**
The GUI should not start.

`sniffnet --help` should show what CLI flags are supported (it should support at least `--help` and `--version`).

`sniffnet --version` should display the current version.

**Desktop (please complete the following information):**
 - OS: Linux
 - Version 1.2.1

Aloso:
This adds `--help`/`-h` and `--version`/`-V` CLI arguments, following the defaults used by [clap](https://docs.rs/clap/latest/clap/). When any of these options is found, we exit with status code 0, which indicates success. When an unknown argument is supplied, we exit with the status code 1. Arguments after `--help`/`--version` are ignored, which I think is not a big deal.

If the plan is to support more CLI arguments, it would make sense to switch to a proper argument parser like clap or [lexopt](https://docs.rs/lexopt/).

Closes #264

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -7,7 +7,7 @@ pub fn parse_cli_args() {
     for arg in args {
         match arg.as_str() {
             "--help" | "-h" => print_help(),
-            "--version" | "-v" | "-V" => print_version(),
+            "--version" | "-v" => print_version(),
             _ => {
                 unknown_argument(&arg);
                 std::process::exit(1);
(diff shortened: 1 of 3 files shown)


GyulyVGC:
I think it's better to not include `clap` just for these two arguments, as you did.
I originally thought to also implement https://github.com/GyulyVGC/sniffnet/issues/243 through CLI arguments, but in the end I think I'll implement it as a feature directly inside the GUI for ease of the users.

Seems fine to me, I will test it as soon as I have time to do so.
Just a little notice is that we could also include `-v` in lowercase (I've seen it more often lowercase wrt uppercase).

Aloso:
@GyulyVGC sure, I'll add `-v` as well.

Regarding the geoip database, another option is to use an environment variable, e.g.

```sh
SNIFFNET_GEOIP_COUNTRY=/path/to/country.mmdb sniffnet
```

which can be added to a shell config file for convenience.

GyulyVGC:
I'm already using [confy](https://docs.rs/confy/latest/confy/) to save configurations for the app (settings info and last successfully sniffed adapter) and I was thinking to adopt the same approach.

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @Aloso for code.

Issue #292, opened 2023-07-02
Missing icon and app name in the dash of Gnome Shell


Digitalone1:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### What's the problem?

![Schermata del 2023-07-02 19-29-19](https://github.com/GyulyVGC/sniffnet/assets/25790525/dd291c0c-1686-44e0-ba35-e9916465a5fd)

As shown in the screenshot. Usually this issue is resolved specifying `StartupWMClass` in the [.desktop](https://specifications.freedesktop.org/desktop-entry-spec/1.0/ar01s05.html) file. Tried it, but nothing changes.

Maybe an issue related to iced library. I don't know. Should be investigated.

### How did you install the app?

downloaded one of the provided packages

### Operating System

Linux

### Additional context

https://archlinux.org/packages/extra/x86_64/sniffnet/

Digitalone1:
Making some researches I found out these:
- https://github.com/rust-windowing/glutin/issues/1067
- https://github.com/rust-windowing/winit/pull/700
- https://developer.gnome.org/documentation/tutorials/application-id.html

TLDR: The app needs an `id` in Wayland. Maybe it's not set and this causes the issue? The question is whether iced has a way to set it.

Digitalone1:
Maybe the `id` [here](https://docs.rs/iced/latest/iced/settings/struct.Settings.html), but I don't know if it's already specified.

GyulyVGC:
Mm, it's not a problem related to iced.

The icon displays just fine in the Debian based and Red Hat based distros (see here the corresponding files: https://github.com/GyulyVGC/sniffnet/tree/main/resources/packaging/linux).

I didn't personally packages it for Arch Linux.
Maybe @orhun, the package maintainer, can tell you more about this, as I think the problem could reside in how the app was packaged for Arch.

Digitalone1:
Did you test on Gnome Wayland?

Digitalone1:
Look, I replaced the content of [this](https://github.com/GyulyVGC/sniffnet/blob/main/resources/packaging/linux/sniffnet.desktop) desktop file (the custom Arch [one](https://gitlab.archlinux.org/archlinux/packaging/packages/sniffnet/-/blob/main/sniffnet.desktop) uses the keyword key and different categories).

Even tried to reboot the system, but the icon is still not showing.

[...]

PR #297, opened 2023-07-04
Update Changelog and add a comment for Settings id field

Digitalone1:
Just a comment for the purpose of the id string. Related to #292 

Another minor thing. I'll point it out here since it's not so important. I noticed that you set 1190x670 as start size, but in another part of the code you maximize on startup. Don't know if it's intended to work that way but, if you need to maximize on start, maybe don't need to specify a start size...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All Sniffnet releases with the relative changes are documented in this file.
 
 ## [UNRELEASED]
 
-- Added app `id` in order to correctly show the icon and app name on Linux Waylend ([#292](https://github.com/GyulyVGC/sniffnet/issues/119))   
+- Added app `id` in order to correctly show the icon and app name on Linux Wayland ([#292](https://github.com/GyulyVGC/sniffnet/issues/119))   
 - Added support for `--help` and `--version` command line arguments ([#272](https://github.com/GyulyVGC/sniffnet/pull/272))
 - Restructured issue templates to let users open issues in a more efficient and effective way ([#285](https://github.com/GyulyVGC/sniffnet/pull/285))
 - Updated French translation to v1.2 ([#279](https://github.com/GyulyVGC/sniffnet/pull/279))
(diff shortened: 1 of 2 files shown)


GyulyVGC:
It's actually made on purpose.
If not specified, iced will use its default size (which I don't remember) and the screen will be maximised anyway an instant later.
Specifying a custom initial size has the benefit of setting it back when double clicking on the top most bar of the app.
I'm not aware of a configuration to immediately set it to the full screen size.
Note that it's not even suggested to set it to the screen resolution, since iced uses logical pixels which are not always equivalent to physical ones.

Digitalone1:
Ok. I updated the changelog.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All Sniffnet releases with the relative changes are documented in this file.
 
 ## [UNRELEASED]
 
+- Added app `id` in order to correctly show the icon and app name on Linux Wayland ([#292](https://github.com/GyulyVGC/sniffnet/issues/119))
```
GyulyVGC, line 8: issue number and link should be consistent :P

src/main.rs
```diff
@@ -101,6 +101,10 @@ pub fn main() -> iced::Result {
 
     print_cli_welcome_message();
 
+    // The id string field is essential for Linux Wayland, especially for correctly showing
+    // the app name and icon on Gnome Shell dash (see issue #292).
+    // Besides the id string should be the same specified for `StartupWMClass` keyword in
+    // `sniffnet.desktop` file for Linux packaging.
```
GyulyVGC, line 107: The comment is fine, I just think that placing it directly above the id field is better

GyulyVGC: Still better if it's a bit shorter

CHANGELOG.md
Digitalone1: Yes, fixed.

src/main.rs
Digitalone1: See if it's good now.

GyulyVGC (approved):

src/main.rs
GyulyVGC: Approved 👍

Issue #262, opened 2023-06-08
Alt+Tab switches tabs


Aloso:
**Describe the bug**
Alt+Tab is used to switch between windows. It should not switch tabs within an application.

**To Reproduce**
1. Open sniffnet
2. Press Alt+Tab to switch to a different window
3. Switch back to sniffnet
4. Notice that a different tab is focused

**Expected behavior**
Alt+Tab should not switch tabs.

**Desktop (please complete the following information):**
 - OS: Linux
 - Version 1.2.1

**Additional context**
Installed with `cargo install sniffnet`

GyulyVGC:
This little bug was introduced just a few days ago in #248.
I will personally fix it before the next release 👍

Aloso:
Oh, I think the bug is two-fold:

- When sniffnet is focused, pressing Alt+Tab immediately switches to the next tab (introduced in v1.2.1)
- When another program is focused, pressing Alt+Tab to switch to sniffnet may also switch to the next tab (existed before)

The 2nd bug seems to only occur when the Tab key is released after the Alt key.

GyulyVGC:
Thanks for the detailed report.
I'll better investigate.

GyulyVGC:
I've restored the original situation (that of v1.2.0, with the difference that now also `esc` and `enter` only apply the shortcut when no modifier at all is pressed).

> * When another program is focused, pressing Alt+Tab to switch to sniffnet may also switch to the next tab (existed before)
> 
> The 2nd bug seems to only occur when the Tab key is released after the Alt key.

I was not able to reproduce this behaviour though.
If Sniffnet is not focused, shouldn't receive commands at all.
The fact that it happens when you release the tab key later suggests that it triggers when no modifiers are pressed, which is what we want (if the app is focused ofc).

GyulyVGC:
`alt`+`tab` should not have an effect now on the app.
Could you try releasing the `tab` slightly before the `alt`?
I guess that the problem is that if you have `tab` pressed (without modifiers) even for a short instant, the app immediately captures it.
Really weird that's happening with the app unfocused though.
On which OS are you running @starccy?
Feel free to provide a short video if you want, so that I can better understand your specific situation.

starccy:
@GyulyVGC

 I am using Kali Linux with Gnome 44 + X11. 

> alt+tab should not have an effect now on the app.

Even if you released `tab` after `alt`?

[...]

PR #298, opened 2023-07-04
fix: ignore tab switch when user focused on window just now

starccy:
related to: #262 

When pressing `Alt` + `Tab` to switch back to the app, the Tab pressed event will be captured immediately, which causes this issue.

I resolve to use the iced window focus event to determine if users switched between the windows and record the timestamp of the focus event meanwhile.
When the `SwitchPage` message is emitted, first check if users just switched windows within a short period. If so, just ignore this request.

### DIFF
--- a/src/gui/app.rs
+++ b/src/gui/app.rs
@@ -104,10 +104,14 @@ impl Application for Sniffer {
 
     fn subscription(&self) -> Subscription<Message> {
         use iced_native::keyboard::{Event, KeyCode, Modifiers};
+        use iced_native::window;
         const NO_MODIFIER: iced_native::keyboard::Modifiers =
             iced_native::keyboard::Modifiers::empty();
         let hot_keys_subscription =
             iced_native::subscription::events_with(|event, _| match event {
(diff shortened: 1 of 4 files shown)


GyulyVGC:
Amazing, I will review and try it in the following days.
Tyvm 🙏

GyulyVGC:
I only simplified something I thought unnecessary, but the overall process is the same and it's good imo.

Before merging I'd like that @Aloso could confirm the problem is solved for him as well.
I still wasn't able to reproduce this problem on my macOS, so having the confirmation from who originally opened the issue would be the best thing.

Aloso:
Thanks! I'll check later today 🙂

Aloso:
@starccy @GyulyVGC I tried this branch and can confirm that it fixes the bug for me (on Arch Linux with KDE Plasma) 🚀

GyulyVGC:
Awesome! Thanks for the confirmation 😄

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @starccy for code.

Issue #288, opened 2023-06-27
Support for ICMP


GyulyVGC:
### Describe the solution you'd like

Many folks requested for ICMP protocol support, which is an important feature that's still missing in the app.
I've decided to finally open an issue since this request was not officially reported yet.
Support for ICMP will be introduced in v1.3.0.

PR #308, opened 2023-07-15
ICMP handling

rrobgill:
Identifies ICMP v4 or v6 packets, and collapses them to a single ICMP transport type. 
Ports are undefined by ICMP, therefore to zero. 
Protocol type set to ICMP to allow filtering.
Issue #288


GyulyVGC:
Thanks so much!
The PR could remain open for a bit, since at least one other release will be published before v1.3.

GyulyVGC:
Closing this PR in favour of a more complete implementation provided in #417

PR #324, opened 2023-07-30
Migration to Iced 0.10

GyulyVGC:
This PR migrates Sniffnet to [Iced 0.10](https://github.com/iced-rs/iced/releases/tag/0.10.0), fixing associated changes and introducing new features available:
- Now the graphical renderer is chosen at runtime and a fallback one (`tiny-skia`) will be used if the default one (`wgpu`) crashes
- Added option to set different types of color gradients for each of the available themes
- Improved tabs and modals layout using the new `BorderRadius` capabilities, that allow to specify a different radius for each of the angles
- Changed scrollable style to make scrollable render properly also with `tiny-skia`
- Unfortunately, Persian translation has been removed since RTL languages are unsupported and may crash the app


GyulyVGC:
@4r7if3x hey man!

Iced has finally published the release featuring a new software renderer as well as font fallback.

- Negative news first:
Unfortunately, I had to comment (80a88717480f83885beb7a15e04f75055559ae69) Persian since the situation for RTL languages got even worse actually (can cause panic in some circumstances and are completely messed up, https://github.com/iced-rs/iced/issues/1981)
- On the bright side:
Now a CPU-only renderer based on tiny-skia is available and is claimed to run "everywhere"; glow has been removed and by default wgpu is still uses, and tiny-skia is used as a fall-back in case of wgpu problems.
tiny-skia is a bit slow in debug mode but works great on my hardware in release mode.
IIRC, you said you had a PC which doesn't support wgpu. Can you confirm it correctly displays everything with glow?
To note that tiny-skia can also be tested by disabling iced default features (that include wgpu)

PR #339, opened 2023-08-13
First-class theming

GyulyVGC:
The concept of [first-class theming](https://github.com/iced-rs/rfcs/blob/master/text/0006-theming.md), a mechanism to make widget theme management more modular and less error-prone, was introduced by iced some months ago in its version 0.5, but I kept delaying the introduction of this feature in Sniffnet because it was something requiring lots of refactoring.

Well… I’ve finally decided to upgrade the theme management strategy of Sniffnet to follow the best practices recommended by iced 🎉 

This has no impact on what the end users see but it makes the code of the GUI more readable and compact.


tarkah:
Awesome! Theming is a great addition. So we created type aliases that we use across the entire app. If you notice, we import these types from our widget module where we've defined the type aliases. 

So any place where we return a specific widget type or need to add concrete type annotations, we ensure to use those type aliased imports that use our internal theme.

PR #341, opened 2023-08-15
Update README.md

bdantas:
Added instructions to install Sniffnet in Tiny Core Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,17 @@ pkgin install sniffnet
 ```
 
 </details>
+<details>
+
+  <summary>on Tiny Core Linux x86_64 port (a.k.a. Core Pure 64)</summary>
+
+You can install Sniffnet from the official repository with:
+
+  ```
+tce-load -wi sniffnet
+```
+
+</details>
 
 ## Required dependencies
 


GyulyVGC:
Amazing, tyvm!

Do you know by chance where TCL users are mostly active?
I'd like to share the availability of Sniffnet, but saw that the Reddit community is pretty inactive.

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @bdantas for doc, platform.

bdantas:
> Amazing, tyvm!
 
It's my pleasure. I like sniffnet and enjoy enriching the TCL repository with new packages.

> Do you know by chance where TCL users are mostly active? I'd like to share the availability of Sniffnet, but saw that the Reddit community is pretty inactive.
 
TCL users and developers are active at https://forum.tinycorelinux.net, but I don't think they like fanfare about new packages.

GyulyVGC:
Got it. Thanks again!

GyulyVGC:
Last clarification: on Wikipedia I see many type of TCL... this package can only be downloaded on CorePure64?
Isn't there a way to make it available for all the TCL types?

Source: https://en.wikipedia.org/wiki/Tiny_Core_Linux

bdantas:
**TL;DR explanation:** 
I, personally, only deal with the x86_64 port ("CorePure64") of TCL. Sniffnet is now in the official x86_64 repository and will work for users using that port--regardless of the installation media that was used ("Core", "Tiny Core", or "Core Plus").

**Detailed explanation:**
The nomenclature for different TCL versions couldn't be more confusing, unfortunately.

At the highest hierarchical level there is Tiny Core Linux (TCL) and there is dCore. The latter is a separate, TCL-inspired OS that imports packages from Debian and Ubuntu. What follows below pertains to TCL only, not to its dCore cousin.

TCL is available as 6 ports for different CPU architectures (aarch64, armv6, armv7, armv7l, x86, and x86_64), as you can see from the way the official repository is structured. Take a look at the official repository for current TCL version (14): http://repo.tinycorelinux.net/14.x. So far, so good.

Now the confusing part.

One confusing thing is that some of the TCL ports have nicknames such as "CorePure64" for the x86_64 port and "piCore" for the armv6 port.

Another confusing thing is that, within a given port, there are several installation media that differ only in the number of included packages (in increasing number of included packages: Core < Tiny Core < Core Plus).

Yet another confusing thing is that some minimalistic users with 64bit machines prefer to use 64bit kernel but 32bit applications from the x86 repository. I believe this setup is called "Core64".

So there you have it :)

bdantas:
The bottomline is that most people using TCL on their desktop or laptop computer are using the x86 or x86_64 port. Based on amount of activity in the x86 and x86_64 repositories, it seems that there are more x86_64 users than x86 users. Therefore, with Sniffnet now being in x86_64 repo, most people running TCL on their computer are now able to install Sniffnet.
 
I'd volunteer to add Sniffnet to the x86 repo as well, but I already maintain about 90 packages in the x86_64 repo and don't want to stretch myself too thin.

GyulyVGC:
What a mess 😂 

Thanks you for the very detailed yet interesting explanation.

Don't worry, I don't want you to take even more responsibility for what concerns Sniffnet and its packaging 

In the past I used to distribute exclusively the x86_64 version of the .deb package as well because of its popularity wrt other architectures, so it's fine as it is now 👍

PR #356, opened 2023-08-26
fix build failure on powerpc64

### LINKED ISSUE #353: Build failure for ppc64

### What's the problem?

Hello. I get an error while has tried to build sniffnet 1.2.2 for ppc64 using ALT Linux Sisyphus. See [logs](https://git.altlinux.org/tasks/327845/build/100/ppc64le/log).  It occurs due to ring which 0.16.2 version [doesn't support](https://github.com/briansmith/ring/issues/1567) this arch. 

### How did you install the app?

built locally by cloning the repository

### Operating System

Linux

### Additional context

Version 1.1.3 has been built for ppc64 successfully. [Logs](https://git.altlinux.org/tasks/archive/done/_317/325478/build/1100/ppc64le/log).

dutyrok:
Disable rustls-tls features of reqwest dependency for ppc64 because of ruslts needs ring, which the latest stable version doesn't support this arch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,6 +717,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "errno"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b30f669a7961ef1631673d2766cc92f52d64f7ef354d4fe0ddfd30ed52f0f4f"
+dependencies = [
+ "errno-dragonfly",
+ "libc",
+ "windows-sys 0.48.0",
+]
+
+[[package]]
 name = "errno-dragonfly"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


GyulyVGC (approved):

GyulyVGC:
So, this cargo.lock is built on a target architecture different from powerpc64 and we should keep this version on the repo.

Did you try building it also on powerpc64 as a confirmation? Is everything fine?

GyulyVGC:
@all-contributors please add @dutyrok for bug.

dutyrok:
> Did you try building it also on powerpc64 as a confirmation? Is everything fine?

Yes, sniffnet has been built on this arch and works (at least in qemu). I put a build log as a proof of build is success.
[sniffnet-1.2.2-ppc64.log](https://github.com/GyulyVGC/sniffnet/files/12447536/sniffnet-1.2.2-ppc64.log)

GyulyVGC:
Hey @dutyrok `reqwest` was updated today and now uses `rustls` by default, which now depends on ring 0.17 that [added support for `powerpc64`](https://github.com/briansmith/ring/issues/1904#issuecomment-1890231721).

For this reason, I've just updated the `reqwest` dependency to use `rustls` also on `powerpc64`.

PR #384, opened 2023-10-10
Update smithay-client-toolkit

AaronDewes:
This is required for supporting KDE Plasma 6.

See also https://github.com/Smithay/client-toolkit/pull/406 and https://github.com/wez/wezterm/issues/3996.


GyulyVGC:
I took advantage of your notice to run a `cargo update` in 7282645.

Thanks very much for pointing out!

PR #397, opened 2023-11-02
Brazilian Portuguese Translation

clr-cera:
# 🇧🇷 Brazilian Portuguese Translation
This PR translates Sniffnet to brazilian portuguese. 

Although european portuguese and brazilian portuguese are very similar, they differ in several ways and what "feels good" in one may not be the same in the other.

### DIFF
--- a/src/gui/pages/settings_language_page.rs
+++ b/src/gui/pages/settings_language_page.rs
@@ -25,15 +25,17 @@ pub fn settings_language_page(sniffer: &Sniffer) -> Container<Message, Renderer<
     let row_language_radio_4 = language_radios(language_active, &Language::ROW4, font);
     let row_language_radio_5 = language_radios(language_active, &Language::ROW5, font);
     let row_language_radio_6 = language_radios(language_active, &Language::ROW6, font);
+    let row_language_radio_7 = language_radios(language_active, &Language::ROW7, font);
     let col_language_radio_all = Column::new()
-        .spacing(12)
+        .spacing(7)
         .align_items(Alignment::Center)
         .push(row_language_radio_1)
         .push(row_language_radio_2)
         .push(row_language_radio_3)
         .push(row_language_radio_4)
         .push(row_language_radio_5)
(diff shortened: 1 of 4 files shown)


GyulyVGC:
I thank you a lot for your contribution, but we already had a discussion about this topic as a PR for Brazilian Portuguese was submitted in #134 and one for European Portuguese was submitted in #136.

In particular, this was the response by @franciscoBSalgueiro who submitted the European version:
> It's the same language, so any portuguese speaking person would be able to read both. That being said, you can definitely tell one was written by a Brazilian and another by a Portuguese. I've seen some apps have both a pt_PT and pt_BR translation, but if you only want to have one, I think any of them would work.

And from that I decided to only include the European version for reasons of easier maintainability.

What do you think?

clr-cera:
Oh sorry, I haven't seen those PRs! 

However as a Brazilian I do think they differ in relevant ways, one sure could read both translations, but there are several different words and habits between them! There are some Portuguese ways of speaking that although are 100% comprehensible, in Brazil they would be unusual or sometimes even wrong.

I saw that European Portuguese is not updated to translations_2.rs, and I don't feel able to complete it as I don't speak this variation of the language and as a consequence would be heavily biased.

In the other hand your decision is completely understandable for practicality and I should indeed have searched for PRs about this.

GyulyVGC:
Don't worry!
Given your considerations, I think I'll include your translation as well.
Even if it will add the overhead of managing one more language, it can be a valuable addition for native Brazilians.

GyulyVGC:
Hey @clr-cera I was about to merge the PR but I noticed that the changes with respect to the European variant are minimal.
At this point I think it's better to only keep one language.
Would you like to open a new PR just to update the original PT translation? (i.e., including only `translations_2.rs` from this PR)

clr-cera:
Sure! I will close this PR and soon I'll send a new one, at the end there will be an European and Brazilian portuguese mix, which I don't think is inherently bad either.

PR #409, opened 2023-12-13
Copy IP addresses to clipboard and add IP address in search parameters inside Inspect page

### LINKED ISSUE #406: Allow functionality to copy from inspect tab (Ctrl+C or CMD+C)

Screenshot 2023-11-29 at 10 58 06

If it allows copy of these useful domains in case someone want to search for more info about it.
this could also be useful to add it in inspect fields whenever possible 

_No response_

GyulyVGC:
This PR implements the possibility to copy IP addresses to the clipboard, to enable users do further analysis about sniffed addresses in an easier way.
IP addresses can now be copied from the popup related to a given entry of the connections table.
It was necessary to manually introduce such an option because the GUI library in use doesn't allow to select text yet.

Furthermore, a new text box has been introduced in Inspect page to allow users filter their connections based on IP addresses values.

https://github.com/GyulyVGC/sniffnet/assets/100347457/58127019-492d-45ce-a1d0-a9fafc19f836

Fixes #406.


itfidele:
@GyulyVGC  This looks so cool, it looks even better than what i was expecting 🎉

PR #419, opened 2023-12-26
Improve custom palettes generation

GyulyVGC:
This PR simplify the user's activity when defining custom color palettes from files.

Before this PR, a user had to specify the following parameters:
- primary
- secondary
- outgoing
- text_headers
- text_body
- starred
- buttons
- round_borders_alpha
- round_containers_alpha
- chart_badge_alpha
- is_nightly

Now, users can simply define a palette specifying the following colors:
- primary
- secondary
- outgoing
- text_headers
- text_body
- starred

...and Sniffnet will automatically infer if the palette is dark or light, determining as a consequence the color of buttons and opacity values.

To achieve this, a struct `PaletteExtension` has been defined; it is computed only once per theme (thanks to the use of `once_cell`) starting from a defined `Palette`.
The struct `PaletteExtension` doesn't only include the parameters that were removed from the user definition of the color scheme, but also two more fields that host the fonts of body text and headers text.

Moreover, the style of the 8 additional themes has been updated to follow the same derivation rules of the styles defined via TOML.


GyulyVGC (commented):
🚢

PR #420, opened 2023-12-27
Graceful termination

GyulyVGC:
Adds support for graceful termination in presence of `SIGINT`, `SIGTERM`, and `SIGHUP`.

Sniffnet needs to save some configurations (settings, window properties, and name of last sniffed device) before closing.

Gracefully terminating the application when the aforementioned signals are received means being able to save such configurations when the user quits Sniffnet in an unconventional way (e.g., by pressing `ctrl` + `C` in terminal, by sending `kill` command, or by closing the terminal that launched the execution).


GyulyVGC (commented):
🚢

PR #421, opened 2023-12-30
Add support for more link types: raw IP and null/loopback

### LINKED ISSUE #30: Fails to capture traffic from most non-physical interfaces
**Describe the bug**

It works for `wlan0` or veth interfaces, but fails for non-Ethernet-like things like `any` or wireguard.

**To Reproduce**

1. Install sniffnet v1.0.0 using `cargo install sniffnet` on Linux Debian 11.
2. Start sniffnet.
3. Choose e.g. `any`.

**Expected behavior**

It considers the same packets as e.g. `tcpdump -i any`.

**Actual behavior**

"No traffic can be observed because the adapter you selected has no active addresses"

or

"No traffic has been observed yet. Waiting for network packets..."

GyulyVGC:
Before this PR, analysing non-Ethernet-based network adapters resulted in failures because packets weren't correctly parsed.

This PR adds support for parsing packets that doesn't start with an Ethernet header.
Now users can analyse more network adapters of their PC, not only the Ethernet-based ones: support for raw IP frames, and null/loopback frames has been added.

Notable examples of usage that are enabled by this PR include the case of VPNs adapters (often based on TUN-like devices), and the case of the loopback interface.

Fixes #30 
Related issue: https://github.com/JulianSchmid/etherparse/issues/78

---

### The strategy

Given an active `Capture`, its link type is determined via [pcap::Capture::get_datalink](https://docs.rs/pcap/latest/pcap/struct.Capture.html#method.get_datalink).

Based on the link type value, packets will be parsed with different strategies:
- if the link type is `RAW`, `IPV4`, or `IPV6`, packets will be parsed starting from their IP headers
- if the link type is `NULL` or `LOOP`, packets will be parsed starting from their IP headers after having verified the validity of the first 4 bytes (`AF_INET` or `AF_INET6` header)
- otherwise, packets will be parsed starting from their Ethernet headers like in the past

Support for other link types may be added in future PRs (see issue #422)

### DIFF
--- a/src/secondary_threads/parse_packets.rs
+++ b/src/secondary_threads/parse_packets.rs
@@ -206,8 +206,6 @@ fn get_sniffable_headers<'a>(
 ) -> Result<PacketHeaders<'a>, ReadError> {
     match is_link_type_confirmed {
         false => {
-            *is_link_type_confirmed = true;
-
             let ethernet_result = PacketHeaders::from_ethernet_slice(packet);
             let ip_result = PacketHeaders::from_ip_slice(packet);
             let null_result = PacketHeaders::from_ip_slice(&packet[4..]);
@@ -218,17 +216,20 @@ fn get_sniffable_headers<'a>(
 
             match (is_ethernet_sniffable, is_ip_sniffable, is_null_sniffable) {
                 (true, _, _) => {
+                    *is_link_type_confirmed = true;
                     *link_type = Linktype::ETHERNET;
                     info_traffic_mutex.lock().unwrap().link_type = Linktype::ETHERNET;
                     ethernet_result
                 }
                 (_, true, _) => {
+                    *is_link_type_confirmed = true;
                     // it could be IPV4 as well as IPV6 but it should be the same
                     *link_type = Linktype::IPV4;
                     info_traffic_mutex.lock().unwrap().link_type = Linktype::IPV4;
                     ip_result
                 }
                 (_, _, true) => {
+                    *is_link_type_confirmed = true;
                     // it could be NULL as well as LOOP but it should be the same
                     *link_type = Linktype::NULL;
                     info_traffic_mutex.lock().unwrap().link_type = Linktype::NULL;
--- a/src/secondary_threads/parse_packets.rs
+++ b/src/secondary_threads/parse_packets.rs
@@ -50,141 +50,145 @@ pub fn parse_packets(
                     return;
                 }
                 if let Ok(headers) = get_sniffable_headers(&packet, my_link_type) {
-                    let mut exchanged_bytes = 0;
-                    let mut mac_addresses = (None, None);
-                    let mut icmp_type = IcmpType::default();
-                    let mut packet_filters_fields = PacketFiltersFields::default();
+                    {
+                        let mut exchanged_bytes = 0;
+                        let mut mac_addresses = (None, None);
+                        let mut icmp_type = IcmpType::default();
+                        let mut packet_filters_fields = PacketFiltersFields::default();
 
-                    let key_option = analyze_headers(
-                        headers,
-                        &mut mac_addresses,
-                        &mut exchanged_bytes,
-                        &mut icmp_type,
-                        &mut packet_filters_fields,
-                    );
-                    if key_option.is_none() {
-                        continue;
-                    }
-
-                    let key = key_option.unwrap();
-                    let application_protocol = get_app_protocol(key.port1, key.port2);
-                    let mut new_info = InfoAddressPortPair::default();
-
-                    let passed_filters = filters.matches(&packet_filters_fields);
-                    if passed_filters {
-                        new_info = modify_or_insert_in_map(
-                            info_traffic_mutex,
-                            &key,
-                            device,
-                            mac_addresses,
-                            icmp_type,
-                            exchanged_bytes,
-                            application_protocol,
+                        let key_option = analyze_headers(
+                            headers,
+                            &mut mac_addresses,
+                            &mut exchanged_bytes,
+                            &mut icmp_type,
+                            &mut packet_filters_fields,
                         );
-                    }
-
-                    let mut info_traffic = info_traffic_mutex
-                        .lock()
-                        .expect("Error acquiring mutex\n\r");
-                    //increment number of sniffed packets and bytes
-                    info_traffic.all_packets += 1;
-                    info_traffic.all_bytes += exchanged_bytes;
-                    // update dropped packets number
-                    if let Ok(stats) = cap.stats() {
-                        info_traffic.dropped_packets = stats.dropped;
-                    }
-
-                    if passed_filters {
-                        info_traffic.add_packet(exchanged_bytes, new_info.traffic_direction);
-
-                        // check the rDNS status of this address and act accordingly
-                        let address_to_lookup =
-                            get_address_to_lookup(&key, new_info.traffic_direction);
-                        let r_dns_already_resolved = info_traffic
-                            .addresses_resolved
-                            .contains_key(&address_to_lookup);
-                        let mut r_dns_waiting_resolution = false;
-                        if !r_dns_already_resolved {
-                            r_dns_waiting_resolution = info_traffic
-                                .addresses_waiting_resolution
-                                .contains_key(&address_to_lookup);
+                        if key_option.is_none() {
+                            continue;
                         }
 
-                        match (r_dns_waiting_resolution, r_dns_already_resolved) {
-                            (false, false) => {
-                                // rDNS not requested yet (first occurrence of this address to lookup)
+                        let key = key_option.unwrap();
+                        let application_protocol = get_app_protocol(key.port1, key.port2);
+                        let mut new_info = InfoAddressPortPair::default();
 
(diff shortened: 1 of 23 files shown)


src/secondary_threads/parse_packets.rs
```diff
@@ -45,152 +50,206 @@ pub fn parse_packets(
...
+                                exchanged_bytes,
+                                new_info.traffic_direction,
+                            ));
                     }
                 }
             }
         }
     }
 }
+
```
JulianSchmid, line 200: Adding a helper function with slice length check and AF number checking (for other comments bellow):

```suggestion
fn try_from_null_frame<'a>(frame: &'a [u8]) -> Result<PacketHeader<'a>, ReadError> {
    if frame.len() < 4 {
        return Err(ReadError::UnexpectedEndOfSlice(packet.len()));
    }
    let is_ip_null_frame_af_num = {
        // based on https://wiki.wireshark.org/NullLoopback.md (2023-12-31)
        fn matches(value: u32) -> bool {
            match value {
                // 2 = IPv4 on all platforms
                // 24, 28, or 30 = IPv6 depending on platform
                2 | 24 | 28 | 30 => true,
                _ => false,
            }
        }
        let s = &frame[..4];
        let b = [s[0], s[1], s[2], s[3]];
        // check both big endian and little endian representations
        // as some OS'es use native endianess and others use big endian
        matches(u32::from_le_bytes(b)) || matches(u32::from_be_bytes(b))
    }
    
    if is_ip_null_frame_af_num {
        PacketHeaders::from_ip_slice(&packet[4..])
    } else {
        Ok(PacketHeaders{
            payload: &[],
            ..Default::default()
        })
    }
}
```

src/secondary_threads/parse_packets.rs
```diff
@@ -45,152 +50,206 @@ pub fn parse_packets(
...
+    packet: &'a Packet,
+    link_type: &mut Linktype,
+    info_traffic_mutex: &Arc<Mutex<InfoTraffic>>,
+    is_link_type_confirmed: &mut bool,
+) -> Result<PacketHeaders<'a>, ReadError> {
+    match is_link_type_confirmed {
+        false => {
+            let ethernet_result = PacketHeaders::from_ethernet_slice(packet);
+            let ip_result = PacketHeaders::from_ip_slice(packet);
+            let null_result = PacketHeaders::from_ip_slice(&packet[4..]);
```
JulianSchmid, line 211: (same comment as bellow)

```suggestion
            let null_result = try_from_null_frame(packet);
```

src/secondary_threads/parse_packets.rs
```diff
@@ -33,6 +33,11 @@ pub fn parse_packets(
 ) {
     let capture_id = *current_capture_id.lock().unwrap();
 
+    // pcap seems to assign ethernet to all interfaces (at least on macOS)...
```
JulianSchmid, line 36: I see a different behavior on my MacOS. If I run

```rust
fn main() {
    // list all of the devices pcap tells us are available
    for device in pcap::Device::list().expect("device lookup failed") {
        println!("Found device! {:?}", device);
        
        let Ok(cap) = pcap::Capture::from_device(device) else {
            println!("    Error creating capture");
            continue;
        };
        let Ok(cap) = cap.immediate_mode(true).open() else {
            println!("    Error opening capture");
            continue;
        };

        {
            let l = cap.get_datalink();
            println!("  Main data link: {:?} {:?} {:?}", l, l.get_name(), l.get_description())
        }

        if let Ok(links) = cap.list_datalinks() {
            for l in links {
                println!("    {:?} {:?} {:?}", l, l.get_name(), l.get_description());
            }
        }
    }
}
```

I get my loopback device correctly reported as "LOOP":

```
...
Found device! Device { name: "lo0", desc: None, addresses: [Address { addr: 127.0.0.1, netmask: Some(255.0.0.0), broadcast_addr: None, dst_addr: None }, Address { addr: ::1, netmask: Some(ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff), broadcast_addr: None, dst_addr: None }, Address { addr: fe80::1, netmask: Some(ffff:ffff:ffff:ffff::), broadcast_addr: None, dst_addr: None }], flags: DeviceFlags { if_flags: LOOPBACK | UP | RUNNING, connection_status: NotApplicable } }
  Main data link: Linktype(0) Ok("NULL") Ok("BSD loopback")
    Linktype(0) Ok("NULL") Ok("BSD loopback")
...
```

Maybe recheck on your system again, then you could potentially remove some logic.

GyulyVGC: Solved this one.
The problem was that I was creating the capture from the device name (`&str`) instead of opening it from the device itself (`pcap::device`).

src/secondary_threads/parse_packets.rs
```diff
@@ -45,152 +50,206 @@ pub fn parse_packets(
...
+                    *link_type = Linktype::NULL;
+                    info_traffic_mutex.lock().unwrap().link_type = Linktype::NULL;
+                    null_result
+                }
+                (false, false, false) => ethernet_result,
+            }
+        }
+        true => match *link_type {
+            Linktype::IPV4 | Linktype::IPV6 => PacketHeaders::from_ip_slice(packet),
+            Linktype::NULL | Linktype::LOOP => PacketHeaders::from_ip_slice(&packet[4..]),
```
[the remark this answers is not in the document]
GyulyVGC: You're absolutely right, thanks 👍

src/secondary_threads/parse_packets.rs
GyulyVGC: Why not returning an `Err` in case `is_ip_null_frame_af_num` is false?

GyulyVGC: I'm not checking errors (just discarding them) so I think it's reasonable to put there one of the `ReadError`s even if not meaningful for this specific case.

GyulyVGC: I think that
``` rust
         Err(ReadError::IoError(std::io::Error::new(
            ErrorKind::InvalidData,
            "Invalid AF_INET / AF_INET6 value",
        )))
```
is the most appropropriate one

src/secondary_threads/parse_packets.rs
```diff
@@ -45,152 +49,188 @@ pub fn parse_packets(
...
         }
     }
 }
+
+fn get_sniffable_headers<'a>(
+    packet: &'a Packet,
+    link_type: Linktype,
+) -> Result<PacketHeaders<'a>, ReadError> {
+    match link_type {
+        Linktype(12) | Linktype::IPV4 | Linktype::IPV6 => PacketHeaders::from_ip_slice(packet),
```
GyulyVGC, line 200: Here I added `LinkType(12)` because doing some tests I discovered that some adapters are tagged with this instead of `IPV4` or `IPV6`.
However a link type of 12 is [not listed anywhere](https://www.tcpdump.org/linktypes.html) so I'm not sure where it's coming from...
Any idea @JulianSchmid?

src/secondary_threads/parse_packets.rs
JulianSchmid: That is also fine. I just had no error in mind to use and just assumed that it would be filtered out by the upper layer anyways.

src/secondary_threads/parse_packets.rs
JulianSchmid: When I have a look at my machine I see some interfaces that are tagged as both "Ethernet" and "12":

```
Found device! Device { name: "ap1", desc: None, addresses: [Address { addr: _, netmask: Some(ffff:ffff:ffff:ffff::), broadcast_addr: None, dst_addr: None }], flags: DeviceFlags { if_flags: UP | RUNNING, connection_status: Disconnected } }
  Main data link: Linktype(1) Ok("EN10MB") Ok("Ethernet")
    Linktype(1) Ok("EN10MB") Ok("Ethernet")
    Linktype(12) Ok("RAW") Ok("Raw IP")
```

So 12 seems to be "RAW IP". But I never did see it as a main data link type only in the list of data links.

I just setup a crate to print all PCAP devices https://github.com/JulianSchmid/print-pcap-devices/tree/v0.1.0 https://crates.io/crates/print-pcap-devices . Maybe have a look what it says there?

GyulyVGC: I already used that script and I obtained similar results.
On Linux TUNs have 12 also as main link type.
Screenshot 2023-12-31 at 14 00 01

I was just wondering why it's not tagged as IPV4 or IPV6 instead, but apparently is just a different way to define the same thing...

JulianSchmid: Ah "TUN" I remember what those are. That are "virtual" network devices that can be used to send & receive network traffic directly on the IP layer. Use-full for VPNs, VMs & so on.

So at the end of the day, yeah from a sniffnets perspective it is the same 😅.

GyulyVGC: Perfect, in fact I tried it and all seems to work properly.

Thank you very much for all the help 🙏 💯

GyulyVGC (commented):
🚢

GyulyVGC:
@all-contributors please add @JulianSchmid for code, ideas.

PR #426, opened 2024-01-07
File dialogs

GyulyVGC:
This PR makes it possible to use native OS file dialogs to select file paths, enabling an easier selection of both custom styles from TOML files and custom DBs from MMDB files.

Since Iced doesn't have a file dialog widget, I decided to use [`rfd`](https://github.com/PolyMeilex/rfd) for this purpose, which should work on all the major operating systems.


GyulyVGC (commented):
🚢

PR #442, opened 2024-01-28
Upgrade inspect table

### LINKED ISSUE #63: Avoid dealignment of report when longer addresses are received

GyulyVGC:
This PR finally adds a proper tabular structure for connections in inspect page (until now, the table columns were kept aligned using a determined amount of whitespaces).

This fixes #63: the new table structure will remain aligned for every possible field value length.

Moreover, new search filters have been introduced (one for each table column) so that it's possible for users to accurately filter connections according to several parameters: source and destination IP, source and destination port, and protocols, in addition to the previously available search filters. 
Note that by default, for ease of use, search filters check for partial matches; if you want strict equality, just add '=' before the desired filter value.

Finally, now it's possible to sort the table also for ascending value of bytes and packets, in addition to the already existing sorting strategy (by default, it's still sorted by most recent).

image

Note how longer values (in this case in the table header) don't cause the table to be dealigned; hovering on them will show the full value.


GyulyVGC (commented):
🚢

PR #446, opened 2024-02-01
Support more upper layer services (many, many, many more)

### LINKED ISSUE #374: [Request] More Port/Application Info

The below provide a lot of info on what application/protocol uses each port.
There is some overlap but, each provides some different info.

/etc/services
https://svn.nmap.org/nmap/nmap-services
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

At the moment there are a lot of ports that are not displaying any info against or appearing in the application panel.

GyulyVGC:
> [!WARNING]
> This PR was replaced by #450

***

As requested in #374 and by multiple other users of the application (e.g., #289, #338), this PR adds support for several new upper layer services in addition to [the 24 already supported application protocols](https://github.com/GyulyVGC/sniffnet?tab=readme-ov-file#supported-application-layer-protocols).

The full list of the new supported services, taken [from Nmap](https://github.com/nmap/nmap/blob/master/nmap-services), is basically the [IANA assigned ports list ](http://www.iana.org/assignments/port-numbers) plus some trojans and worms.
It contains **6438 different services/protocols/trojans/worms**.

While I could use a file-lookup approach to resolve the name of a service given a port number and the transport protocol, I decided to create [a short bash script](https://github.com/GyulyVGC/sniffnet/pull/446/files#diff-bcc505a8dfead95eba0bda7ffc5a295b8a7f7b0a37b3e912708e184055283fb1) that automatically curls the full list of services and generates a Rust file containing a single (very long) `match` expression wrapped in a new method.
This strategy may seem a little unconventional but should guarantee `un-matchable` performances (pun intended) in the port-to-service mapping procedure.

Since apparently there isn't a way to autofix `clippy::match_same_arms`, the bash script also takes care of merging duplicate branches of the `match`.

Fixes #374


services.sh
```diff
@@ -0,0 +1,30 @@
...
+# The source for port-to-service mappings is provided by nmap
+# Available at: https://raw.githubusercontent.com/nmap/nmap/master/nmap-services
+
+OUT=./src/networking/services.rs
+
+{
+  printf '//! This file is generated automatically via ./services.sh\n\n'
+  printf 'use crate::networking::types::protocol::Protocol;\n\n'
+  printf '#[allow(clippy::too_many_lines, clippy::unnested_or_patterns)]\n'
+  printf 'pub fn get_service(port_info: (u16, Protocol)) -> String {\n'
```
mina86, line 13:
```suggestion
  printf 'pub fn get_service(port_info: (u16, Protocol)) -> Option<&'static str> {\n'
```

so the caller can decide how they want to represent unknown services and whether they need owned String.  If you care about performance you should avoid String allocations.

bradengroom (approved):
lgtm

yawaramin:
If you do this, you'll have to cut a new release every time you want to change even a single character of any of the service definitions. It will quickly become unsustainable.

GyulyVGC:
> If you do this, you'll have to cut a new release every time you want to change even a single character of any of the service definitions. It will quickly become unsustainable.

I see your point.

Anyway, I don't judge it a priority to update this service list.
The most common services assignments are stable, well-known, and long lasting conventions maintained directly by IANA.

GyulyVGC:
After [having long discussed](https://www.reddit.com/r/rust/comments/1agnojr/im_about_to_release_in_production_a_7k_lines_long/?utm_source=share&utm_medium=web2x&context=3) this PR with the Rust community of Reddit, I came to the conclusion that a better strategy is to generate a compile-time, static, perfect hash map for the network services.
The new approach still guarantees high performances and avoids the complexities of this PR.

New PR: #450.

PR #450, opened 2024-02-06
Support more upper layer services (many, many, many more)

### LINKED ISSUE #374: [Request] More Port/Application Info

GyulyVGC:
> [!NOTE]
> This PR replaces the original #446

***

As requested in #374 and by multiple other users of the application (e.g., #289, #338), this PR adds support for several new upper layer services in addition to the 24 already supported application layer protocols.

The full list of the new supported services, taken [from Nmap](https://github.com/nmap/nmap/blob/master/nmap-services), is basically the [IANA assigned ports list ](http://www.iana.org/assignments/port-numbers) plus some trojans and worms.
It contains **6438 different services/protocols/trojans/worms**.

Services are stored as a compile-time, static, perfect hash map generated in `build.rs` with the help of [rust-phf](https://github.com/rust-phf/rust-phf).
This strategy guarantees high performances while avoiding the complexities of #446.

Fixes #374


GyulyVGC (commented):
🚢

PR #452, opened 2024-02-14
Sort options for Host & Service columns in overview page

GyulyVGC:
Until now, network hosts and services in overview page were always sorted by descending value of cumulative transmitted data (bytes or packets).
This was pretty limiting, primarily because the user couldn't easily see the most recent items involved in network traffic exchanges.

This PR implements the possibility to set a sorting strategy for both the network host and service lists.

By default, items are now sorted by most recent timestamp of last data exchange.
With a simple click users can change this behaviour to sort items by ascending/descending value of cumulative transmitted data (bytes or packets).


GyulyVGC (commented):
🚢

PR #461, opened 2024-02-21
Use smoother curves to interpolate traffic chart data points

GyulyVGC:
This PR changes the **interpolation method** used for data points represented in the traffic chart.

Instead of linear interpolation, now a **cosine-like spline** is used.
This results in a smoother and overall better-looking traffic chart (see pictures below).

***

<div align=center>

<h3> Before </h3>

<h3> After </h3>

</div>


GyulyVGC (commented):
🚢

PR #470, opened 2024-03-09
Migrate to Iced 0.12

GyulyVGC:
Update the code to use Iced latest version.

- updated `Modal` widget
- removed generic arguments from `Renderer`
- improved button shadows
- color gradients are now computed in Oklab space
- improved tooltips of buttons linking to external web pages
- simplified `Length`s invocations
- improved toggler's border appearance


GyulyVGC (commented):
🚢

PR #493, opened 2024-04-08
Update Italian language

bovirus:
@GyulyVGC 

Please merge. Thanks.

### DIFF
--- a/src/translations/translations.rs
+++ b/src/translations/translations.rs
@@ -8,7 +8,7 @@ use crate::StyleType;
 pub fn choose_adapters_translation(language: Language) -> Text<'static, StyleType> {
     Text::new(match language {
         Language::EN => "Select network adapter to inspect",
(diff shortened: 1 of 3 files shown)


GyulyVGC:
Hi!
Thanks for your attempt but I'm Italian myself and I don't find your modifications to be improving / impactful enough to justify a PR.

bovirus:
@GyulyVGC 

I know very well that you are Italian (I can see by your name).

You used terms that are quite not stanbdard.
Ex "Adattatore di rete" is not usual 
The most used and known term is "Scheda di rete"

You used a lot of time the article when in english is not used and is not sense in Italia.
"File di database" is correct but unusal.
The most used term is "File database".

GyulyVGC:
I think neither your nor mine translations are _inherently wrong_ / _more correct_ than the other, and that's just about it.
I mostly actually prefer my translations, but I agree with you that in many cases other people may find yours more suitable, it's just a matter of habit.

PR #495, opened 2024-04-08
new german translations

luca3s:
added the new german translations. While I was at it I adjusted some old ones.


GyulyVGC (approved):

GyulyVGC:
Thanks very much for the new translations and also for the fixes! 🙌

PR #504, opened 2024-04-13
add Japanese translation

hirotake111:
This is a PR to add Japanese translation to v1.3 (Related issue: #60)

As you may know some Japanese character are missing because of the same reason as [this](https://github.com/GyulyVGC/sniffnet/issues/60#issuecomment-1678166954). So please fill in them.


GyulyVGC:
Sure I'll take care of the missing code points. Thanks 🎉

GyulyVGC (approved):

PR #523, opened 2024-04-29
clippy: apply missing_const_for_fn lint

CosminPerRam:
This PR applies the [missing_const_for_fn](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_const_for_fn) lint and also adds it to the CI.

Technically the majority of them should be already optimized away, but if they can be const, should they be const?

I also found 3 instances of translations functions having per item `.to_string()` instead of on the entire match block and a single translation function call that was cloned unnecessarily.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -59,7 +59,7 @@ jobs:
 
       - name: clippy
         if:  matrix.os != 'windows' || github.event_name != 'pull_request'
-        run: cargo clippy -- -D warnings
+        run: cargo clippy -- -W clippy::all -D clippy::missing_const_for_fn -D warnings
 
       - name: test
         if: matrix.os != 'windows' || github.event_name != 'pull_request'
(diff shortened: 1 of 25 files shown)


GyulyVGC:
As mentioned in the same [lint page](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_const_for_fn) you linked:

> Const functions are currently still being worked on, with some features only being available on nightly. This lint does not consider all edge cases currently and the suggestions may be incorrect if you are using this lint on stable.

In general I think that's better to rely on the lints included by default in `clippy warnings`, since they are the ones considered to be more stable and impactful.

I'm also used to run quite often `clippy pedantic`, and not even that includes this specific lint.

I prefer waiting until this is stabilised.

Thanks anyway for your effort!

CosminPerRam:
Well, we don't have to add that lint to CI, but other than that, the changes are there (more constness).

GyulyVGC:
Yeah, as you say I don't think it'd be an issue to merge these changes, but since the official docs themselves warn that suggestions may be incorrect, I prefer waiting the stabilisation of the lint to be sure to apply it only where really necessary.

PR #524, opened 2024-04-29
Reduce usage of `String` in translations code.

CosminPerRam:
This is a part of #523 (the non-lint part of it).
This PR removes an unnecessary `to_string` call on an already `String` value and replaces per-item `.to_string()` with an entire match block `.to_string()`.


GyulyVGC:
Thanks!

While you are at it, would you mind updating all the translations methods to return an `&'static str`?

I think there are some cases where I actually don't need them to be strings, so this should be more efficient.

CosminPerRam:
Although there was a mix usage of `&'static str`, `String` and `Text<'static, StyleType>`, I ought here to replace only the non-parameter-composing `String` instances as replacing `Text` ones would require bigger changes (that can be done in another PR).

GyulyVGC:
> Although there is a mix usage of `&'static str`, `String` and `Text<'static, StyleType>`, I ought here to replace only the non-parameter-composing `String` instances as replacing `Text` ones would require bigger changes (that can be done in another PR).

I agree, and I'm fine with this.

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @CosminPerRam for code.

PR #558, opened 2024-07-07
Add support for IPinfo MMDB format

### LINKED ISSUE #533: Support IPinfo MMDB database format
### What's the problem?

In version 1.3.0 I introduced the possibility to select custom MMDB databases for country and ASN resolutions (you can find out more about it [here](https://github.com/GyulyVGC/sniffnet/wiki/Custom-databases)).

However, I noticed that [IPinfo](https://ipinfo.io) (one of the best IP data providers globally) uses a slight different format for such databases, which is currently unsupported by Sniffnet (because the application expects databases to be in the canonical MMDB format).

### Additional context

The goal is to support both the canonical format and IPinfo's custom MMDB format, so that users of the app will be able to select IPinfo high quality databases.

If you are interested about why IPinfo uses a different format, I suggest you to read [this](https://ipinfo.io/blog/migrating-from-maxmind-to-ipinfo/) article.

I also want to thank a lot @abdullahdevrel who gave me amazing insights about this topic and wrote a detailed [community post](https://community.ipinfo.io/t/using-ipinfos-mmdb-database-with-rust/5587) about how to interface with IPinfo's MMDBs in Rust.

Related issue: https://github.com/oschwald/maxminddb-rust/issues/82

GyulyVGC:
This PR adds support for IPinfo MMDB database format as described in #533.

Now both the "standard" MMDB files and IPinfo's ones can be successfully imported and used with Sniffnet.

This was achieved introducing new enums whose variants correspond to the possible different representations of database entries.
In this way, the result of the lookup against the MMDB file will be deserialised to the proper enum variant.

Fixes #533.


GyulyVGC (commented):
🚢

PR #577, opened 2024-08-02
[I18N: VN] Add Vietnamese translation  for all 3 versions

HUNG-rushb:
Add Vietnamese translation  for all 3 versions


src/translations/translations.rs
```diff
@@ -445,6 +460,7 @@ pub fn network_adapter_translation(language: Language) -> &'static str {
         Language::FI => "Verkkosovitin",
         Language::JA => "ネットワーク アダプター",
         Language::UZ => "Tarmoq adapteri",
+        Language::VN => "Network adapter",
```
GyulyVGC, line 463: missing translation?

HUNG-rushb: I think that keep it like that is fine, translate it into Vietnamese is kinda long and might be misunderstood, confusing.  We  call it Network adapter directly most of the times. :)

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @HUNG-rushb for translation.

PR #641, opened 2024-11-22
Improve time values on the horizontal axis of the chart

### LINKED ISSUE #619: Improve time values on the horizontal axis of the chart

Currently, the app displays the number of seconds since the beginning of the capture on the horizontal axis of the chart in the _Overview_ page.

For better readability, we should display elapsed time as `hh:mm:ss`

_Note: since only the past 30 seconds of activity are displayed, showing timestamps would be meaningless._

_No response_

GyulyVGC:
See #619 for more information. Fixes #619.


GyulyVGC (commented):
🚢

PR #645, opened 2024-11-26
Redirect `stderr` and `stdout` to file on Windows releases

### LINKED ISSUE #578: Impossible to see `stderr` and `stdout` of release builds on Windows

### What's the problem?

I set the configuration attribute [`windows-subsystem="windows"`](https://rust-lang.github.io/rfcs/1665-windows-subsystem.html#new-crate-attribute) for Windows release builds to avoid having a terminal running alongside Sniffnet when opening the app by clicking on its icon.

However, this prevents users to see `stderr` and `stdout` even when launching the app from terminal; in case of errors, this doesn't allow to see a trace of the problem, making debug more difficult.

### Additional context

A possible solution that comes to my mind is to redirect `stdout` and `stderr` to a new file in the app's configuration directory.

GyulyVGC:
Redirect `stderr` and `stdout` to file on Windows releases (fixes #578).

It also adds a new CLI option `--logs` to open the file where the logs are written (only for Windows targets).


GyulyVGC (commented):
🚢

PR #659, opened 2024-12-18
Network host filters autocompletion via a dropdown menu

### LINKED ISSUE #354: `ComboBox` widget for search filters

Screenshot 2023-08-24 at 10 31 03 PM

Instead of typing, I was wondering if user could select from the existing domains fetched from the backend(over view page) so user could select directly

This would save time to go back to overview page and to copy the exact domain name and could select from a dropdown box.

I would like to get started with drop-down but if a dropdown feature exists already I would refer it directly. Let me know the checks! 

Thanks!

GyulyVGC:
This PR implements network host filters autocompletion via a dropdown menu.

This way, it's now possible to preview featured countries, domains, and ASNs while typing in the respective text input in the _Inspect_ page.

https://github.com/user-attachments/assets/e060c70b-2a44-44bc-a2dc-ff7dc62fc68b

Fixes #354.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All Sniffnet releases with the relative changes are documented in this file.
 
 ## [UNRELEASED]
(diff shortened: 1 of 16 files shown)


GyulyVGC (commented):
🚢

PR #678, opened 2025-01-12
Provide details about unassigned IPs (bogons)

### LINKED ISSUE #209: Bogon tag (and detailed address type identification)
### Discussed in https://github.com/GyulyVGC/sniffnet/discussions/208

<div type='discussions-op-text'>

<sup>Originally posted by **nitefood** May 21, 2023</sup>
Just like for multicast addresses (which you already identify and tag accordingly), I'd like to suggest a _bogon_ tag, along with more detailed address type identification (e.g. RFC1918 private space, or RFC6598 CGN space), to get rid of many unnecessary "unknown location" occurrances.

For example, [these](https://github.com/nitefood/asn/blob/a19482aab9a024c7baa425fdd0b899fbd0bfcb09/asn#L556-L567) are some regex patterns I used for a similar purpose.</div>

GyulyVGC:
Until today, Sniffnet tagged as 'Unknown location' all IP addresses for which it wasn't possible to determine the origin.
This PR adds support for tagging _bogon_ IPv4 and IPv6 addresses.

> [!NOTE]
> Some IP addresses and IP ranges are reserved for special use, such as for local or private networks, and should not appear on the public internet.
> These reserved ranges, along with other IP ranges that haven't yet been allocated and therefore also shouldn't appear on the public internet are sometimes known as bogons.

***

Sniffnet will now represent with a _lock_ icon that a connection was established with a bogon IP address, and hovering on it will show additional details about such IP.

Screenshot 2025-01-19 at 17 53 20

Fixes #209.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All Sniffnet releases with the relative changes are documented in this file.
 
+## [UNRELEASED]
+- Identify and tag unassigned/reserved "bogon" IP addresses ([#678](https://github.com/GyulyVGC/sniffnet/pull/678) — fixes [#209](https://github.com/GyulyVGC/sniffnet/issues/209))
+
 ## [1.3.2] - 2025-01-06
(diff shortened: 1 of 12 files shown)


GyulyVGC (commented):
🚢

PR #692, opened 2025-02-01
Add Ukrainian translation for version 1.3.

Kapelianovych:
I have updated the Ukrainian translation to support 1.3 version and fixed some existing translations (mostly making it more polite and removing obvious mistakes).


GyulyVGC (approved):

GyulyVGC:
It doesn't happen often that a contributor also takes care of updating the changelog.
Thanks so much for your time!

GyulyVGC:
@all-contributors please add @Kapelianovych for translation.

src/translations/translations.rs
```diff
@@ -990,7 +990,7 @@ pub fn packets_chart_translation(language: Language) -> &'static str {
         Language::ES => "paquetes por segundo",
         Language::PL => "pakiety na sekundę",
         Language::DE => "Pakete pro Sekunde",
-        Language::UK => "пакети(ів) на секунду",
```
Kapelianovych: I personally wasn't able to check where this translation is used, but there is a special type of conjugation in Ukrainian language which depends on a preceding number. For example, if there is a `2 packets per second`, it translates `2 пакети на секунду`, but `11 packets per second`, translates to `11 пакетів на секунду`. Considering that I added parentheses with alternative ending.

GyulyVGC: Hey, I actually removed the `(ів)` from all the cases since I think it's fine to also use the singular. By the way that translation is used on the traffic chart heading when you switch to packets as data representation.

Kapelianovych: I see. Well, let it be like this for now. Maybe in future i18n feature can be improved to infer a context for a translation.

PR #714, opened 2025-02-13
Grouped country mappings

ZeroDot1:


GyulyVGC:
Hey! Nice to hear back from you.

Thank you, but I don't see how this could impact the logic of the match statement, and I don't want to risk breaking existing logic if not necessary.
Also I prefer having them sorted alphabetically.

Issue #715, opened 2025-02-14
Add icon to titlebar


jermanuts:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Describe the solution you'd like

An icon in the titlebar

![Image](https://github.com/user-attachments/assets/16b9a6be-feb5-4707-ac4e-fae69945ebdd)

https://docs.iced.rs/iced/window/struct.Icon.html

Relevant issue https://github.com/iced-rs/iced/issues/587

### Is your feature request related to a problem?

yes, there's no icon in titlebar for Windows

GyulyVGC:
Thanks for the suggestion

GyulyVGC:
Yes, I just have to verify whether it correctly sets the icon.
I remember I tried in the past without success, but probably this was fixed in newer iced versions.

GyulyVGC:
Awesome, thanks. Feel free to send a PR 😁

PR #717, opened 2025-02-15
Add the Titlebaricon solve #715

ZeroDot1:


GyulyVGC:
Problems:
- the icon you're using is the old one
- we shouldn't include the icon file in the `src` folder
- calling `expect` can make the program crash

I've fixed this myself in #719

Issue #553, opened 2024-07-04
ENH: Add link to wiki/user manual in the footer off the GUI


corneliusroemer:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Describe the solution you'd like

Add a link to the official user manual/wiki/docs in the footer of the GUI

Pointing to this: https://github.com/GyulyVGC/sniffnet/wiki

### Is your feature request related to a problem?

I'm always frustrated when I can't quickly find the docs/help section of a program I'm using

corneliusroemer:
I only now after exploring the app further discovered that you link to the docs in "settings > general help" - that's great, but a bit hard to find. I wouldn't expect the link to docs be a general setting :) You have plenty of footer space, use it!

GyulyVGC:
It makes sense!

P.S.: I just wanted to let you know that you probably uploaded the wrong video/picture not only in this issue but also in other issues you previously opened.

GyulyVGC:
@all-contributors please add @corneliusroemer for ideas.

corneliusroemer:
Ah thanks! I'm sorry I opened so many issues, I hope it's not overwhelming you, sometimes I just have a run of ideas 🙈

GyulyVGC:
No problem at all — the more issues/ideas/feedbacks the better it is, because it allows me to get inspirations to always improve!

PR #733, opened 2025-02-21
Add link to wiki in GUI footer

cxt240:
Just fiddling around with rust and saw your repo and this issue (https://github.com/GyulyVGC/sniffnet/issues/553).

Change I made should look like this: 
image

Please let me know if there are any issues with where I placed the link (was unsure if it should've been a button instead, but thought text was alright) or if you wanted to proceed along another path. Thanks!


GyulyVGC:
Hey, thanks for your effort!

Adding that link is something I'd like to do myself because it will also require removing the current link from the settings page, plus adding other additional links in the footer.
I want them to be buttons and in an odd number to have the GitHub one in the center, so it's something that I have to still think quite a bit.

PR #756, opened 2025-03-13
Donut chart reporting traffic statistics

### LINKED ISSUE #687: Add donut chart reporting total data statistics
Instead of having total data statistics in written form (see picture below), it'd be nicer to have them as donut charts reporting:
- incoming data
- outgoing data
- filtered out data
- dropped data

Image

Iced doesn't support pie/donut charts out of the box, so there is the need to manually draw and update it using a `canvas` widget.

GyulyVGC:
This PR adds a donut chart reporting overall traffic statistics: the amount of data incoming / outgoing / filterd out / dropped.

I believe this is a nice way to visualise data in a simplified way, and is also more visually appealing!

| Before | After |
|---|---|
|  |  | v1.4.0

***

https://github.com/user-attachments/assets/f80fe6d9-6284-444d-925a-29fbaf865041

***

Fixes #687.


GyulyVGC (commented):
🚢

Issue #680, opened 2025-01-14
Add support for ARP


GyulyVGC:
Considered the new capabilities of `etherparse` v0.17 (#679), we should consider adding support for the [Address Resolution Protocol](https://en.wikipedia.org/wiki/Address_Resolution_Protocol) in Sniffnet.

shu-kitamura:
Hello, @GyulyVGC  
Nice to meet you.

I'm shu-kitamura.
I'm interested in this issue.
I'd like to try to add this feature.

I modified some functions so that I can capture ARP packets in my local environment.
I confirmed that I can see ARP packets on the screen as shown below.

<img width="947" alt="Image" src="https://github.com/user-attachments/assets/bdf4f843-26ca-462a-a706-559b7ff0a9d2" />

![Image](https://github.com/user-attachments/assets/a95c6d40-fea2-4d10-8c5a-db6a20917989)

As I understand it, ARP packet support is as shown in the image. Is it correct?

This is my first time contributing to OSS, so I'm sure there are many things I don't understand.

GyulyVGC:
Hi @shu-kitamura, thanks for your interest!

Yes there is probably something to improve but we'll discuss about it in the PR eventually.

For now, I just want to say that I'm not sure it's completely correct to include ARP as "protocol" since it's on a different layer of the ISO/OSI stack with respect to TCP/UDP/ICMP.

However it doesn't even make sense to include it as an IP version, so yeah probably the most meaningful thing is to categorize it as protocol.

This uncertainty is also the reason why I was hesitant to implement this feature.

GyulyVGC:
What I'd also like for ARP is to have the list of addresses that were object of the resolution: similarly to ICMP for which I report the list of messages, it'd be nice to add the list of addresses that were resolved by each ARP connection.

shu-kitamura:
Please let me confirm.  
Are you saying that, just as the ICMP output shows a list of source/destination IP addresses, the ARP output should also display a similar list that clearly indicates which IP addresses were resolved?

GyulyVGC:
I'm referring to the information that is available when you click on one of the rows in the inspect page table.
For ICMP there is a section reporting the list of message types part of that connection.
I believe we should report something also for ARP, like maybe the number of requests VS responses message, or other details that we can extract from the etherparse ARP header.

PR #759, opened 2025-03-16
Add ARP support

shu-kitamura:
Hello!
I'm shu-kitamura.

I took on the challenge #680.

ARP is listed in the protocol list in the menu.

image

ARP is also shown in the Inspect table.

image

When I click on the ARP row in the Inspect page table, the type of ARP operation is displayed.

image

### DIFF
--- a/src/networking/manage_packets.rs
+++ b/src/networking/manage_packets.rs
@@ -153,7 +153,6 @@ fn analyze_transport_header(
     protocol: &mut Protocol,
     icmp_type: &mut IcmpType,
 ) -> bool {
-    println!("{:?}", transport_header);
     match transport_header {
         Some(TransportHeader::Udp(udp_header)) => {
             *port1 = Some(udp_header.source_port);
--- a/src/networking/manage_packets.rs
+++ b/src/networking/manage_packets.rs
@@ -1288,7 +1288,8 @@ mod tests {
                         Protocol::TCP => Service::Name("netstat"),
                         Protocol::UDP => Service::Unknown,
                         Protocol::ICMP => panic!(),
-                        Protocol::ARP => panic!(),                    }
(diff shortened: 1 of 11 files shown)


GyulyVGC:
Thank you so much!

I'm pretty busy these days, but I'll review the PR as soon as I get some time. 🙌

src/networking/manage_packets.rs
```diff
@@ -119,6 +130,20 @@ fn analyze_network_header(
             *exchanged_bytes += u128::from(40 + ipv6header.payload_length);
             true
         }
+        Some(NetHeaders::Arp(arp_packet)) => {
+            *network_protocol = match arp_packet.proto_addr_type {
+                EtherType::IPV4 => IpVersion::IPv4,
+                EtherType::IPV6 => IpVersion::IPv6,
+                _ => return false,
+            };
+            let src: [u8; 4] = arp_packet.sender_protocol_addr().try_into().unwrap();
```
GyulyVGC, line 139: From a quick look I think this is going to panic in presence of IPv6 ARP. If the protocol addresses aren't of the expected length OR they don't are IPv4/IPv6 addresses, we want to ignore that packet.

shu-kitamura: Thanks for your quick check.

About IPv6, I overlooked...

In commit https://github.com/GyulyVGC/sniffnet/pull/759/commits/35d42d0d531af3f33e735b91c617795d8d6d666d, I fixed to not panic when the address is IPv6.
Also, if the protocol address is not the expected length or is not an IPv4/IPv6 address, the packet will be ignored.

GyulyVGC (approved):
:ship:

GyulyVGC:
@all-contributors please add @shu-kitamura for code.

PR #777, opened 2025-04-13
Apply new notification thresholds only after timeout expires

vcrn:
Changes to notifications settings are stored temporarily in temp_notifications field. These are saved to configs only after a timeout has expired. The checking of this timeout and whether there's any new notifications settings to store are done in `TickInit` and `TickRun` arms of `fn update()`.


vcrn:
I tried some different approaches but finally decided to put this one up for review since this implementation is the most straightforward one to me, and the one that is the most consistent with existing design.

Note that it's not just the notifications thresholds that are instantly applied, but all notifications settings.

What do you think? Some things that bother is that the checking whether to update the notifications settings is done for every Tick, and that the very first if-state `if !self.timing_events.was_just_notifications_edit()` will be true for a majority of the time that Sniffernet is running.

Also, when looking at the docs for `TickInit` and `TickRun`, it seems like `TickInit` should be every 5 seconds and `TickRun` every second, but they both use `PERIOD_TICK`, which is 1 second. Is that something that should be changed?

By the way, I noticed in

src/gui/sniffer.rs
```diff
@@ -635,6 +644,19 @@ impl Sniffer {
         self.configs.lock().unwrap().settings.scale_factor
     }
 
+    /// Updates notification settings if it hasn't been edited for a while
+    fn update_notification_settings(&mut self) {
+        // Ignore if just edited
+        if !self.timing_events.was_just_notifications_edit() {
+            if let Some(temp_notifications) = self.temp_notifications.take() {
```
vcrn, line 651:
```
if let Some(temp_notifications: Notifications) = self.temp_notifications {
    if !self.timing_events.was_just_notifications_edit() {
        self.temp_notifications = None;
```
Turning the logic around and adding one extra line would make this code slightly more effective, since 
`if !self.timing_events.was_just_notifications_edit()`
 is true for the majority of the time that the app is running. This isn't the case however for 
 `if let Some(temp_notifications: Notifications) = self.temp_notifications`.

GyulyVGC:
Some remarks from a quick review.

- [ ] 1. this behavior should only apply to threshold changes

- [ ] 2. no need to update notification settings in `TickInit` since notifications won't be received in the initial page of the app; you can just call the update to notifications inside `Sniffer::refresh_data`

- [x] 3. no need to store the configurations as part of notifications update; in fact, changes to the configuration file are only done during app shutdown (I already fixed this point with my latest commit)

src/gui/sniffer.rs
GyulyVGC: I wouldn't be too worried about this since it only happens once per second and it only consists in updating the value of a `Mutex`, there's no big overhead.
However, yeah, make it to do the least number of checks possible.

vcrn:
Thank you for the fast review! I'll fix the remaining point you brought up in the coming days and fix the failing test

GyulyVGC:
Haven't re-reviewed the code yet, but I notice an issue by simply testing via running Sniffnet: if I click on the checkbox to enable/disable a specific notification kind, the corresponding notifications are enabled/disables with some latency... hence I suspect that enabling/disabling is still subject to the timeout.

vcrn:
Thanks! I'll get it fixed and get back to you. I'll see if I can expand the testing to check for this scenario

vcrn:
If the user changes the `ByteMultiple` of `BytesNotification`, do you think that should be applied immediately?

GyulyVGC:
> If the user changes the `ByteMultiple` of `BytesNotification`, do you think that should be applied immediately?

No. The Byte multiplier is part of the threshold and is set via the text input, so its modification has to be subject to timeout.

vcrn:
I've gone with the other approach of storing the temporary thresholds, in a field of type `Option<(PacketNotification, BytesNotification)>`.

GyulyVGC:
Is the PR ready for review?

vcrn:
Yes, ready for review :)

src/gui/sniffer.rs
```diff
@@ -340,8 +345,18 @@ impl Sniffer {
...
             }
-            Message::UpdateNotificationSettings(value, emit_sound) => {
-                self.update_notification_settings(value, emit_sound);
+            Message::UpdateNotificationSettings(notification, emit_sound) => {
+                // Don't update adjustments to thresholds immediately:
+                // that is, sound and toggling thresholds on/off should be applied immediately
+                // Threshold adjustments are saved in `self.temp_thresholds` and then applied
+                // after timeout, but for simplicity, this field is always updated to whatever the
+                // latest change has been.
+                self.update_sound(notification, emit_sound);
```
GyulyVGC, line 354: Do we really need to have all these different methods to update notification values? Can't we merge all of them into one that updates everything except thresholds to make the code more compact?

src/gui/sniffer.rs
```diff
@@ -127,6 +129,8 @@ pub struct Sniffer {
     pub id: Option<Id>,
     /// Host data for filter dropdowns (comboboxes)
     pub host_data_states: HostDataStates,
+    /// Temporary storage for thresholds while editing.
+    pub temp_thresholds: Option<(PacketsNotification, BytesNotification)>,
```
GyulyVGC, line 133: I suggest this to be part of the field TimingEvents::threshold_adjust like (std::time::Instant, Option<(PacketsNotification, BytesNotification)>).
This way you can update it when you call threshold_adjust_now, and you can make was_threshold_adjust return an Option instead of bool.

src/gui/sniffer.rs
GyulyVGC: Also, I think these could be an impl of Notifications structs rather than Sniffer struct.

vcrn: Sure! I tried out different approaches, and went this since although it's quiet verbose, it's very explicit what's going on. Having it in one method looked quiet convoluted in the end when I tried it, but I'll give it another go, I'll probably have some new ideas of how to make it cleaner :)

src/gui/sniffer.rs
vcrn: Good idea, I'll implement this!

GyulyVGC:
Hey @vcrn how're you doing?

I just wanted you to know that in a few days I'll merge #795 into main.
Since that is a huge PR and would cause many conflicts, I'd like to have this one finished and merged first if possible.

vcrn:
Thanks for the heads-up! I'll try to get something pushed by the end of the day

src/gui/sniffer.rs
vcrn: I've tried having this an an impl of Notifications instead, but I believe it will make things more complicated. The majority of the code in these methods are to unpack and compare notifications settings and finally potentially apply them immediately. If we put it in `impl Notifications` we won't be able to apply them, which in my mind would cause an extra layer somewhere else in the code. What do you think?

vcrn:
I have something in place that is a bit rough but it does what it's supposed to do. Since this latest round is more concerning design, I'm considering slightly polishing what I have and pushing it to hear your thoughts on these changes. What do you think?

src/gui/sniffer.rs
GyulyVGC: Then you can leave it as Sniffer's impl but try to make it just one method that's less verbose

GyulyVGC:
IIRC the last time I tested the most recent commit, it was already working fine from a functional perspective.
So yeah, the main concern was about making it more compact since I think there was too much code for what we're trying to do.

vcrn:
Not sure why clippy failed in the pipeline for MacOS but not the others (Linux at least fine when I run it locally, but the build was cancelled in the pipeline, I guess since MacOS failed before that).

Anyway, clippy for MacOS complains about some code I haven't touched, so you better be the judge of whether it should be changed or ignored.

Regarding my latest change, I created one big method that handles all the logic of updating notifications settings and temporary thresholds. Macros could probably decrease the lines of code, but I tried to keep repetition and abstraction layers to a minimum to get your feedback on which way we should go. 

I'll also add some comments here on GitHub for parts that I think need some extra attention.

src/gui/types/timing_events.rs
```diff
@@ -53,6 +65,31 @@ impl TimingEvents {
...
+    pub fn threshold_adjust_now(
+        &mut self,
+        temp_thresholds: (PacketsNotification, BytesNotification),
+    ) {
+        self.threshold_adjust.0 = std::time::Instant::now();
+        self.threshold_adjust.1 = Some(temp_thresholds);
+    }
+
+    /// If timeout has expired, take temporary thresholds
+    pub fn threshold_adjust_expired_take(
```
vcrn, line 78: Pretty bad name, but I wanted it to be clear that it mutates the value. Also, the logic is changed, since we want a value to returned/taken if enough time has passed: compared with previous version, where `true` was returned if not enough time had passed.

GyulyVGC:
Thanks will check this out.
Don't worry about clippy, I'll fix it. By the way, it's due to the new clippy lints recently added in the latest version of Rust.

src/gui/types/timing_events.rs
GyulyVGC: I'd call it `was_just_threshold_adjust_take`.

vcrn:
Yeah, after updating my toolchain it's failing on Linux as well. A bit odd that it isn't failing for Windows though, but maybe it isn't up to date

src/gui/types/timing_events.rs
GyulyVGC: and turn the logic the way round if possible, otherwise leave it as it is

GyulyVGC:
On Windows the CI/CD is skipped in PRs since they need a secret environment variable to be set at the repo level (used to download Npcap).

src/gui/types/timing_events.rs
vcrn: I don't believe we'll be able to change the behavior of "if timeout has expired, return taken value, else None". If this is the case, I can add a doc string explaining how the method works.

Also, could potentially go with the name `was_not_just_threshold_adjust_take`, to try and point out the somewhat reversed behavior compared with the other methods since to most people `false` ~ `None`, `true` ~ `Some`. At the cost of making the name more convoluted.

GyulyVGC: I'll think to it and will personally take care of it during the review :+1:

GyulyVGC (approved):

GyulyVGC:
Finally merged 🎉 

Thanks mate, I hope I wasn't too pedantic during the process 🤣 

@all-contributors please add @vcrn for code

vcrn:
> Finally merged 🎉
> 
> Thanks mate, I hope I wasn't too pedantic during the process 🤣
> 
> @all-contributors please add @vcrn for code

Don't worry, you weren't too pedantic, you we're great! 🙂 Really enjoyed this PR and learned some stuff as well.

I'll be on the look out for other issues I can help out with. Thanks again for a great app and for all your work!

PR #784, opened 2025-04-17
Error handling to reduce the number of possible crash occurrences

GyulyVGC:
Properly handle errors when possible, instead of calling `unwrap`/`expect`.

A new trait `ErrorLogger` has been introduced to log errors to `stderr` in production, allowing to get information about the error kind and its location in the code without panicking (in development builds it will also panic, to easily allow spotting errors when coding).


GyulyVGC (commented):
:ship:

Issue #788, opened 2025-04-18
Some interactive elements provide a very subtle focus indicator when hovering over them with the mouse


GyulyVGC:
### This is part of #760.

This wants to track point 6. from #760: "Some interactive elements provide a very subtle focus indicator when hovering over them with the mouse"

Issue #760, opened 2025-03-17
Integrate feedbacks from NLnet's Accessibility Audit


GyulyVGC:
Some months ago, [Sniffnet was selected to receive support from NLnet via the Next Generation Internet program](https://sniffnet.net/news/ngi-program/).

As part of the program, Sniffnet received a complete Accessibility Audit.

![Image](https://github.com/user-attachments/assets/394df6fb-1d32-493a-89f2-9f9a4a83b795)

jaskij:
Just to throw in my two cents, since I just tried out Sniffnet: keyboard navigation is absolutely essential not only for accessibility, but also to accelerate the workflow. If I'm switching between a terminal, my IDE, and Wireshark, testing things out, my hands never leave the keyboard. Sniffnet makes that impossible.

PR #789, opened 2025-04-18
Improve color contrast on hover

GyulyVGC:
Improve color contrast on hover of some UI elements as suggested by the Accessibility Audit.

It was pretty straightforward to identify the elements signaled as problematic since they were all buttons of the same kind (`BorderedRoundSelected`).

Fixes https://github.com/GyulyVGC/sniffnet/issues/788.

See https://github.com/GyulyVGC/sniffnet/issues/760 for more info about other Accessibility problems.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -72,16 +72,16 @@ jobs:
         with:
           targets: ${{ matrix.target }}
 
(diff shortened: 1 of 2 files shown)


GyulyVGC (commented):
🚢

PR #806, opened 2025-05-12
Use asynchronous channels to update app state from backend

GyulyVGC:
Until now, some of the app's state fields (especially those related to the parsing packets backend) were wrapped into `Mutex`es to let the secondary threads update them remotely.

This is against Iced (and ELM) principles for which it's not suggested having a shared mutable state: the only way to update the app state should be via `Message`s received by the app's `update` method itself.

For this purpose, asynchronous channels were introduce to update the state by sending `Message`s from the backend when appropriate.

These changes also enables producing new tailored UI updates needed when parsing packets from a PCAP file, and this is the reason why this PR will be merged into #795.


GyulyVGC (commented):
🚢

PR #833, opened 2025-06-27
Add german v4 translations

luca3s:
Hi,
just saw your release announcement, thought i could do this quickly.

Where is "excluded" used? I couldn't find it. If you could tell me how i can get to that menu or give me a screenshot that would really help me decide how it should be translated.

cc https://github.com/GyulyVGC/sniffnet/issues/60


GyulyVGC:
Thanks so much for asking.

It's used in the donut chart when there is some filter set, to indicate the amount of data that is "filtered out".

Like in here: #756

inkreasing:
Ok, i now made the translations specific to filters, it sounds a lot better like this, but if you ever decide to include something else in there it needs to be changed.

I also changed the "dropped" translation from plural to singular, so that all 4 types are singular.

GyulyVGC (approved):

PR #842, opened 2025-06-30
Update translations_4.rs

Charpy:
Update french translation.
Thx for 1.4.0 !


GyulyVGC:
Thanks 🙏

GyulyVGC:
@Charpy there are two translations missing (for bits and pause)

GyulyVGC:
No worries, I noticed that someone else submitted another update for French so I'll use that (#864)

Thanks anyway :+1:

Charpy:
Hi ! As it's the same word in french, I didn't translated "Bits" and "pause". I was thinking it will just use the default string, so no need to write more code.. Sry for confusion.

GyulyVGC:
No problem. I noticed it once I saw the other guy's French translation. I still like to have them even if they're the same because once all the languages complete the translations, the default case is removed.

Anyway no worries 🙌

Issue #252, opened 2023-06-06
App Packaging


GyulyVGC:
(Still incomplete) TODO list about packaging problems/features to be solved/implemented for the various operating systems.

### Windows

- There is a command-line window opening and closing quickly at the end of the _Windows_ setup wizard since no other _WixToolset_ feature for running commands quietly worked as expected.
- The sample graphics used for the _Windows_ installer should be modified. Since transparency is supported, I suggest the dialog background be limited to the left side only to follow the standards and fix the visual issues there. The same can apply to the banner image, traditionally only an icon was shown on the right side of it.

### macOS

- The wrapper script to open the app on macOS should be changed. Known issues: #197, #198, #226

### Linux

- AppImage, Flatpak, and other kinds of distribution methods.

GyulyVGC:
Feel free to list other problems/features that come to your mind and I'll add them @4r7if3x

tmn007:
Could be worth getting an SSL certificate to sign the Windows package. Otherwise Windows Defender gets upset that it is an unknown publisher.

GyulyVGC:
> Could be worth getting an SSL certificate to sign the Windows package. Otherwise Windows Defender gets upset that it is an unknown publisher.

Yep, absolutely. I immediately add it to the set of features.
And similarly should also be done for macOS...

4r7if3x:
@GyulyVGC You'd need a registered company and enlist it with D&B to obtain a _Code Signing Certificate_. Besides, you can buy them cheaper from resellers such as [GoGetSSL](https://www.gogetssl.com/code-signing-ssl/).

GyulyVGC:
Dang, I was hoping for a for a free solution but it may be utopian.

auronandace:
This may be an option for you: https://github.com/axodotdev/cargo-dist

myselfop143:
#252

celtic-coder:
> Dang, I was hoping for a for a free solution but it may be utopian.

Hi Giuliano (@GyulyVGC),

Perhaps you can check out [SignPath for Open Source projects](https://about.signpath.io/product/open-source)?

Kind Regards,
Liam

GyulyVGC:
Thanks, will check!

AlleM43:
I've been looking at packaging for Flatpak and AppImage for the last couple days.

[...]

PR #859, opened 2025-07-07
Add AppImage Packaging to CI

AlleM43:
AppImage packaging using pkg2appimage to convert the Debian package into an AppImage as part of the packaging workflow.

As mentioned in #252

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -81,14 +81,13 @@ jobs:
 
       - name: Test (release mode)
         if: matrix.os == 'macos' || matrix.os == 'ubuntu' || matrix.os == 'windows' && matrix.arch == 'amd64'
-        run: cargo test --release --verbose -- --nocapture &&
+        run: |
+          cargo test --release --verbose -- --nocapture &&
+          cargo clean
 
       - name: Install Cross
         if: matrix.os == 'ubuntu'
         run: cargo install cross --git https://github.com/cross-rs/cross
-
-      - name: Clean
-        run: cargo clean
 
       - name: Build binary (Linux)
         if: matrix.os == 'ubuntu'
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -177,7 +177,6 @@ jobs:
           - arch: armhf
             appimgarch: armhf
 
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -195,7 +194,7 @@ jobs:
         shell: bash
         run: git clone https://github.com/AppImageCommunity/pkg2appimage.git
 
-      - name: Replace placeholders. pkg2appimage bundled implementation of apt update does not allow download of packages for different architectures by default. Override this.
+      - name: Replace placeholders
         shell: bash
         run: |
           sed -i -e 's/REPLACE_TAG/Sniffnet_LinuxDEB_${{ matrix.arch }}.deb/g' resources/packaging/linux/AppImage/sniffnet.yml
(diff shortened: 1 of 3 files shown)


GyulyVGC:
Thank you 🙏 

Will review and test this as soon as I get enough time.

AlleM43:
Alright. I had some issues getting the existing build workflow running without secrets, but the second to last commit in the unsquashed branch (https://github.com/AlleM43/sniffnet/tree/d2577161117d8361c483347430ad3ea1238f3744) is what ended up running the AppImage build to completion. It should work without those modifications as long as the deb builds.

GyulyVGC:
I noticed you're packaging it only for `amd64`... I'm wondering how AppImages are usually packed for different arch?
Would it be possible to also support at least `aarch64` or `arm64` or also 32-bit architectures? Would it make sense?

AlleM43:
I haven't really looked into that, partially because the only system I have with a DE installed is amd64, and partially because I thought it best to get amd64 working first and try to expand compatibility later. I can take a look at it though.

GyulyVGC:
Thanks. 

Some other observations:
- `sniffnet.yml` should be under the `resources/packaging/linux` folder
- make sure the produced artifact name matches the style of other existing artefacts (`Sniffnet_LinuxAppImage_arch.AppImage`)
- if it's possible, execute a test workflow on dispatch from your repo so we can test the produced artifact

AlleM43:
I've moved the yaml recipe and changed the artifact name. 

https://github.com/AlleM43/sniffnet/actions/runs/16672113220 produced AppImages for amd64, i386, arm64, and armhf. I haven't tested running them, but I tested amd64 and i386 appimages built locally with the same recipe, and checked all executables and libraries in each AppDir to ensure they were for the correct platfor

I also changed cargo clean to be a separate step in order to deal with an issue I was having with cross. (https://github.com/cross-rs/cross/issues/724)

GyulyVGC:
Thank you.

> https://github.com/AlleM43/sniffnet/actions/runs/16672113220 

Do you have any idea why the dmg and msi fail?

> produced AppImages for amd64, i386, arm64, and armhf. I haven't tested running them, but I tested amd64 and i386 appimages built locally with the same recipe, and checked all executables and libraries in each AppDir to ensure they were for the correct platfor

Awesome, I'll also test at least the amd64.
If you could also directly test such artifacts it'd amazing.

> I also changed cargo clean to be a separate step in order to deal with an issue I was having with cross. ([cross-rs/cross#724](https://github.com/cross-rs/cross/issues/724))

I added that cargo clean exactly because I was having issues with cross IIRC... 🤨

AlleM43:
I disabled Windows and MacOS builds in my testing runs because Windows builds failed because of a lack of a secret and MacOS builds take too many GitHub Actions minutes. The one in the PR branch has those reenabled.

The cargo clean only ran when the tests were ran, which I also disabled to save time. Thus, I moved it to a separate step, to ensure it runs every time.

GyulyVGC:
Oh ok, then I guess I can move it back to the original place.

Again thanks so much for your time 🙌

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @AlleM43 for platform.

GyulyVGC:
I've just tested the amd64 and it perfectly works.
Wow. AppImages are cool actually.

GyulyVGC:
I just had a doubt @AlleM43 

Are we sure libpcap is the only needed dependency? See https://github.com/GyulyVGC/sniffnet/wiki/Required-dependencies

It may work on our systems if we already have the other dependencies installed, but not sure if it works in general...

AlleM43:
The other dependencies are part of those assumed by the pkg2appimage devs to be present on every system (see https://github.com/AppImageCommunity/pkg2appimage/blob/master/excludedeblist) and would thus not be installed anyway.

GyulyVGC:
Perfect 👌

GyulyVGC:
Hey @AlleM43 some users reported a problem with the libpcap dependency when installing the AppImage.

Could you please check this thread: https://www.reddit.com/r/linux/s/x8lRJeJXCv

PR #901, opened 2025-07-25
doc: Add links to translated README versions

dowithless:
Added language selection links to the README for easier access to translated versions: German, Spanish, French, Japanese, Korean, Portuguese, Russian, and Chinese.

The updated links can be previewed in my forked repository: https://github.com/dowithless/sniffnet/tree/patch-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,18 @@ Translated in:<br>
 🇨🇳 🇩🇪 🇫🇷 🇷🇺 🇵🇹 🇪🇦 🇮🇹 🇵🇱 [+&nbsp;14&nbsp;more&nbsp;languages](https://github.com/GyulyVGC/sniffnet/issues/60)
 </div>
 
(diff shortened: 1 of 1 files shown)


GyulyVGC:
Thanks for your time, but I'm not totally convinced this is a so useful addition.

It feels like this job is already done pretty well by the default translator on the web browser.

dowithless:
> Thanks for your time, but I'm not totally convinced this is a so useful addition.
> 
> It feels like this job is already done pretty well by the default translator on the web browser.

Thank you for your feedback! The translations here are processed using AI technology. Based on comparative tests, I believe they deliver better readability and fluency than default browser translations.

GyulyVGC:
Another thing that I don't like very much is that users are redirected to an external website from GitHub.

I'll consider this maybe in the future, but for the moment I prefer keeping things as they are 👍

Issue #898, opened 2025-07-24
Wrapper script for macOS: use a relative path to the executable instead of an absolute path


GyulyVGC:
When installing Sniffnet with the provided Disk Image, it requires the app to be placed in the Applications folder in order to work.

This is because on macOS Sniffnet is launched via a [wrapper script](https://github.com/GyulyVGC/sniffnet/blob/main/resources/packaging/macos/wrapper.sh) that references the executable absolute path.
This issue proposes to launch it with a relative path instead so that we can avoid issues like #888.

Problems in fixing this issue:
- we cannot use `$PWD` since it contains the directory where the script is launched, not the script directory
- I almost got it working using `DIR=$(cd "$(dirname "$0")"; pwd -P)`, but then for some reason `$DIR` is not correctly expanded in the osascript "do shell script"

InfraWhisperer:
Hey @GyulyVGC, Please look at my changes to fix this. Thanks!

PR #948, opened 2025-08-31
fix wrapper script to use relative path on macos

InfraWhisperer:
## Summary

Fixes the macOS wrapper script to use relative paths instead of absolute paths, allowing the Sniffnet app to run from any location without requiring installation in the Applications folder.

## Changes

1. Modified wrapper.sh to dynamically detect script directory using $(cd "$(dirname "$0")" && pwd)
2. Changed from hardcoded /*/Sniffnet.app/Contents/MacOS/sniffnet to relative path resolution
3. Used realpath to properly expand the path before passing to osascript to avoid variable expansion issues

## Test plan

1. Verified path resolution works correctly from different directories
2. Confirmed osascript command formation is valid
3. Tested that the script no longer depends on specific installation location

Fixes: #898

### DIFF
--- a/resources/packaging/macos/wrapper.sh
+++ b/resources/packaging/macos/wrapper.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env zsh
-SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
 SNIFFNET_PATH="$SCRIPT_DIR/sniffnet"
 osascript -e "do shell script \"'$SNIFFNET_PATH' >/dev/null 2>&1 &\" with prompt \"Comfortably monitor your Internet traffic.\" with administrator privileges"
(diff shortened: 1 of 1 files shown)


GyulyVGC:
Thanks for your attempt.
However, the `realpath` command seems to not be installed by default on all the macOS versions.
In fact I get this error: `command not found: realpath`.

InfraWhisperer:
> realpath

Sorry, my bad, I assumed realpath exists. I removed the usage of realpath and tested again. Please let me know if this works in your env too.

GyulyVGC:
Nice, it now seems to work! Maybe it's worth it to use `pwd -P` to be sure that symbolic links are resolved.

GyulyVGC:
It's also interesting to notice how your use of single & double quotes is finally expanding the variable, which is what I was failing to do.

I'm wondering whether we can be sure that the variable is expanded correctly in all the environments.

GyulyVGC (approved):
:ship:

GyulyVGC:
@all-contributors please add @InfraWhisperer for platform.

PR #971, opened 2025-09-17
Fix AppImage error with `libpcap`

AlleM43:
https://github.com/GyulyVGC/sniffnet/pull/859#issuecomment-3303347936 Seems to have been caused by Debian 13 changing the name of the libpcap package. I've updated the manifest to use the new name.

### DIFF
--- a/resources/packaging/linux/AppImage/sniffnet.yml
+++ b/resources/packaging/linux/AppImage/sniffnet.yml
@@ -1,8 +1,8 @@
 app: Sniffnet
 ingredients:
-  dist: bookworm
+  dist: trixie
   packages:
-    - libpcap0.8
+    - libpcap0.8t64
   sources:
     - deb https://deb.debian.org/debian stable main
   debs:
--- a/resources/packaging/linux/AppImage/sniffnet.yml
+++ b/resources/packaging/linux/AppImage/sniffnet.yml
@@ -1,9 +1,9 @@
 app: Sniffnet
 ingredients:
-  dist: trixie
+  dist: buster
   packages:
-    - libpcap0.8t64
+    - libpcap0.8
   sources:
-    - deb https://deb.debian.org/debian stable main
+    - deb https://archive.debian.org/debian buster main
   debs:
     - /target/REPLACE_TAG


GyulyVGC:
Will this work with older distributions though? 🤔

GyulyVGC:
And why was it happening even if we were using bookworm?

https://packages.debian.org/search?searchon=sourcenames&keywords=libpcap

AlleM43:
It was happening because the manifest was pulling the package from the repository for the latest version of Debian. I could've modified to keep it pulling from the bookworm repository, but since the .deb is now being built for trixie I decided to update that line and keep it pulling from stable.

AlleM43:
> Will this work with older distributions though? 🤔

I did not consider this, though. libpcap from Trixie requires glibc 2.38. Seeing as the main `sniffnet` binary requires glibc 2.29, I'll be changing it to a version compatible as far back as possible

AlleM43:
Changed it to use Debian 10 libpcap. It would be possible to use Debian 11, but that would cause breakage in August next year when it's archived.

GyulyVGC:
Thanks, I prefer this solution.
Why not keeping `bookworm` at this point?

I'm also wondering whether I should also fix something in the .deb package building and the Docker package (they're both built on a ubuntu-latest runner) 🤔

AlleM43:
Bookworm libpcap requires glibc 2.33, which would have prevented running it on, among others, Debian 11 and Ubuntu 20.04.

GyulyVGC:
What about using ubuntu-latest for building the AppImage as well? Is that possible? So we're consistent with how I'm building the other packages.

AlleM43:
Please elaborate. In what way would ubuntu-latest be used in the AppImage build, beyond being the runner?

GyulyVGC:
Using it as the dist in the ingredients section. But I'm not really sure if it's possible and meaningful.

AlleM43:
That would cause the same problem as using Debian 13, with the libpcap requiring too new versions of glibc.

GyulyVGC:
Makes sense. 
Before merging I'd like to replicate the issue they pointed out to verify that this PR fixes it, but unluckily I'm not able to reproduce the problem.
The original AppImage built on bookworm with libpcap0.8 just works fine for me on my Debian:

image

EDIT: ok, I'm unsure why it's working on Debian, but I was able to finally replicate the issue on Fedora.

AlleM43:
If you remove the libpcap package from your host system you should be able to replicate the issue, or you can run the AppImage with the flag `--appimage-extract` and verify that the folder that's extracted doesn't contain libpcap.

GyulyVGC (approved):

GyulyVGC:
Works fine now!!
This changes are now on main, and the release was updated to include the newly built AppImages.

Thanks so much for your responsiveness and availability.

PR #999, opened 2025-10-31
Fix #639: disable background tooltips

### LINKED ISSUE #639: Background tooltips are visible on hover even when a modal is open in the foreground
When a modal (e.g., a setting page) is open in the foreground, background tooltips are visible when hovering on the corresponding area.

This bug was introduced during the [migration to iced 0.13](#618) and is a regression due to iced itself.

simaosanguinho:
This PR addresses issue #639 .

Before, when a modal was opened, the tooltips in the background would still be visible:
unsolved

And after being solved here is the result:
solved

## Solution
I created a wrapper around the iced Tooltip widget, that given a control parameter (`enabled`) either gives out a tooltip like before or a transparent one which has no visual effect, as shown in the gif above.

All tooltips were changed to the new version to maintain consistency and I used the Builder design pattern to make the tooltip creation as flexible as possible.

Let me know what you think and what else could be changed regarding this issue.

### DIFF
--- a/src/countries/country_utils.rs
+++ b/src/countries/country_utils.rs
@@ -1,5 +1,5 @@
 use iced::Font;
-use iced::widget::Tooltip;
+use iced::{Element};
 use iced::widget::svg::Handle;
 use iced::widget::tooltip::Position;
 use iced::widget::{Svg, Text};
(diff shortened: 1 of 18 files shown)


GyulyVGC:
Thanks for the PR, but I believe that passing the 'show_tooltip' parameter through so many places in the code isn't a good idea, especially considering that iced 0.14 will be soon release and will automatically address this issue.

At this point I think it's better to just wait for the next version of iced.

PR #1002, opened 2025-11-06
Add welcome screen

GyulyVGC:
After more than 3 years... finally adding an animated welcome screen to the app 🙈 
The same animation, reversed and slightly shorter, is shown when quitting the app.

https://github.com/user-attachments/assets/ab4eeab6-4fa6-4250-9a8e-5d1b2b46ae2d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All Sniffnet releases with the relative changes are documented in this file.
 
+## [UNRELEASED]
+- Add animated welcome screen ([#1002](https://github.com/GyulyVGC/sniffnet/pull/1002))
+
 ## [1.4.2] - 2025-11-04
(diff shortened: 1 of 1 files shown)


src/gui/pages/welcome_page.rs
```diff
@@ -0,0 +1,55 @@
...
+    };
+
+    let text = Text::new(match x {
+        0..=3 | 20.. => "",
+        4 => "S",
+        5 => "Sn",
+        6 => "Sni",
+        7 => "Snif",
+        8 => "Sniff",
+        9 => "Sniffn",
```
daveth3t3chg33k: Do not hardcode the variable name extract it to a global file and call it

GyulyVGC (commented):
🚢

Issue #1052, opened 2026-01-09
Add Tailscale service (port 41641)


QazCetelic:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Describe the solution you'd like

Tailscale uses UDP on port 41641 for peer-to-peer connections

### Is your feature request related to a problem?

2nd top service is unknown.

<img width="728" height="248" alt="Image" src="https://github.com/user-attachments/assets/e7c4bc71-0f78-4cc1-b402-8b4654d15c8c" />

GyulyVGC:
The services detected by Sniffnet are the ones recognized by Nmap, and our service list is automatically generated from theirs.

What I can suggest is opening an issue about this on their repo: https://github.com/nmap/nmap

PR #1053, opened 2026-01-09
Add Tailscale to service recognition

QazCetelic:
- #1052

I wasn't able to test it due to dependency issues on OpenSUSE. I hope you can run `cargo test` for me.


GyulyVGC:
As I wrote in #1052:

> The services detected by Sniffnet are the ones recognized by Nmap, and our service list is automatically generated from theirs.
> 
> What I can suggest is opening an issue about this on their repo: https://github.com/nmap/nmap

PR #1055, opened 2026-01-22
Support custom IP blacklists

### LINKED ISSUE #729: Custom IP blacklists
Allow users to import a custom IP blacklist.

When a connection with a blacklisted IP is established, the user should be notified, and the connection should be highlighted in some way.

This is a first step in the direction of making it possible to identify suspicious/malicious traffic.

_Originally posted by @ZeroDot1 in https://github.com/GyulyVGC/sniffnet/issues/685#issuecomment-2661026741_

GyulyVGC:
This PR adds the possibility for users to select a custom IP blacklist from a file.

A new notification kind has been added to be warned when network traffic is exchanged with a blacklisted IP, and a new filter is available in the _Inspect_ page to show only blacklisted connections.

https://github.com/user-attachments/assets/e5bdd4d1-8596-45ff-b399-3dff013befc4

Fixes #729.

### DIFF
--- a/resources/test/ip_blacklist_invalid.txt
+++ b/resources/test/ip_blacklist_invalid.txt
@@ -1,0 +1,6 @@
+Lorem ipsum dolor
+7.7.7.999
+invalid_ip_address
+12.12.1
+300.300.300.300
+example.com
--- a/resources/test/ip_blacklist_valid.txt
+++ b/resources/test/ip_blacklist_valid.txt
@@ -1,0 +1,10 @@
+# this only has 4 valid IPs, but it's still valid!
+  8.8.8.8
+
+ehi
+ciao
+
+fe80::99
+1.2.3.255
+::123
+...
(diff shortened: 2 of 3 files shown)


src/translations/translations_5.rs
```diff
@@ -29,3 +29,27 @@ pub fn remote_notifications_translation(language: Language) -> &'static str {
...
+        Language::EN => "IP blacklist",
+        Language::IT => "Blacklist IP",
+        _ => "IP blacklist",
+    }
+}
+
+pub fn blacklisted_transmitted_translation(language: Language) -> &'static str {
+    match language {
+        Language::EN => "New data exchanged from a blacklisted IP",
+        Language::IT => "Nuovi dati scambiati dai un IP in blacklist",
```
GyulyVGC:
```suggestion
        Language::IT => "Nuovi dati scambiati da un IP in blacklist",
```

PR #1076, opened 2026-02-17
Add Brazilian Portuguese (PT_BR) language support

josealmada:
Add Brazilian Portuguese as a new language option, separate from European Portuguese (PT).

Note:** I am a native Brazilian Portuguese speaker. While adding PT_BR, I also corrected a few European Portuguese (PT) translations that seemed inaccurate. Those PT changes should ideally be reviewed by a Portuguese native speaker.


GyulyVGC:
Hi @josealmada thank you.

I remember in the past someone started adding PT-BR and then we decided to merge the two into a single language because they were very similar.
Can you confirm it or you think there are notable differences? 
Can you also check that PT is actually the European one for all the sentences since I think some PT-BR could also be mixed up there.

In any case, since I plan to introduce more sentences that will require a translation before v1.5.0, I'll wait to merge this PR and will tag you once we're almost there.

josealmada:
They’re very similar. I believe most Brazilians would understand it even without a pt-BR version, as I didn’t notice anything that could cause confusion.

If you’re keeping only one variant, I’d suggest the following adjustments to the PT version:
- "Dropped" (in dropped packets) → use "Descartados" instead of "Perdidos", which implies "lost."
- only_top_30_items → the current sentence is incorrect. Suggested: "Apenas os 30 primeiros itens são apresentados aqui."
- source → "Origem" works better for both variants.
- file → European Portuguese typically uses "Ficheiro", while Brazilian Portuguese uses "Arquivo."

Sounds good — I’ll be happy to help with the new sentences once you add them.

GyulyVGC:
Given they are so similar, I believe it'd be better if you just submit an improvement and fixes to the existing one probably

GyulyVGC:
We're finally there!

Please pull the latest changes on the main branch (you'll find the new sentences in `translations_5.rs`).

Thank you in advance @josealmada  

Also please remember:

> Given they are so similar, I believe it'd be better if you just submit an improvement and fixes to the existing one probably

Issue #1069, opened 2026-02-14
Add support for additional protocols


GyulyVGC:
Currently Sniffnet supports the following protocols:
- ARP
- ICMP
- TCP
- UDP

It'd be a nice plus to add support for:
- IGMP
- SCTP

mbe57:
Is there a way to identify and report torrent traffic, despite ports being a moving target ...

GyulyVGC:
Unfortunately no.

It only makes sense for Sniffnet to associate known ports to services, and the convention is maintained by IANA.

Sniffnet is just using the [nmap](https://github.com/nmap/nmap) services list.

PR #1093, opened 2026-03-06
Add IGMP and SCTP support

Elyrial:
extend the Protocol enum with IGMP and SCTP variants and update analyze_transport_header to detect them from the raw ip payload when etherparse doesn't recognize the transport layer. IGMP gets no ports, SCTP extracts src/dst ports from the first 4 bytes of its common header. also hide the service field for IGMP in the connection details page since it has no ports

---

I did some manual testing as well, I generated both protocol types locally and confirmed they got picked up correclty.

For IGMP, joined multicast group  `239.0.0.1` via a raw UDP socket -> shows up with no port fields and the service field is hidden.

For SCTP, used ncat to send a message to a listener on port 12345 -> ports parse correctly from the common header.

Worth noting you need to capture on `lo` for SCTP loopback traffic, not your network interface.

Closes: #1069

### DIFF
--- a/src/networking/manage_packets.rs
+++ b/src/networking/manage_packets.rs
@@ -559,7 +559,7 @@ mod tests {
     use crate::Protocol;
     use crate::Service;
     use crate::networking::manage_packets::{
-        get_service, get_traffic_direction, get_traffic_type, is_local_connection,
+        analyze_headers, get_service, get_traffic_direction, get_traffic_type, is_local_connection,
         mac_from_dec_to_hex,
     };
     use crate::networking::types::address_port_pair::AddressPortPair;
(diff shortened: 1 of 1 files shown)


GyulyVGC:
Thanks for the PR, even if this probably won't be evaluated in the short term due to me giving precedence to other features 👍

GyulyVGC:
@Elyrial after thinking about this and briefly reviewing your PR, it'd be definitely better to send this PR directly to [etherparse](https://github.com/JulianSchmid/etherparse) so that other people can benefit from this and Sniffnet can automatically reuse it.

I've also opened two issues there: 
- SCTP: https://github.com/JulianSchmid/etherparse/issues/133
- IGMP: https://github.com/JulianSchmid/etherparse/issues/134

Elyrial:
@GyulyVGC sure, that makes more sense, I could implement it once the etherparse project actually confirms they want it.

GyulyVGC:
I believe they would definitely be interested in supporting them.

GyulyVGC:
cc @JulianSchmid

JulianSchmid:
Hi @Elyrial ,

I would merge an implementation if it is solid.

But I cannot guarantee how much time it will take until I merge a PR. I currently don't have a lot of time.

But I would welcome it.

Greets
Julian

Issue #1096, opened 2026-03-07
Enhance and extend Inspect page


GyulyVGC:
Currently the _Inspect_ page of the app shows a single table grouping items by their network 5-tuple (source and destination IP, source and destination port, protocol).

While working on program identification (#1056) and starting to think how to enhance favourites (#839) I came across a major limitation: Sniffnet already captures and stores a bunch of data, but at the moment we're not optimally showing all the correlations and details of such data.

This is mostly because the _Inspect_ page table has no column for fields related with network hosts and programs: a user can filter table entries based on such parameters using the text inputs, but to actually see them in the search results one has to click and expand each result individually.

Some examples:
- there's no immediate way to see all the network hosts (country, domain, ASN) associated with a given program
- there's no immediate way to see all the programs associated with a network host (country, domain, ASN)
- there's no immediate way to see network hosts (country, domain, ASN) and programs associated with blacklisted IPs
- filtering by favourite doesn't make it clear whether the results are favourite hosts, services, or programs

A possibility to fix this could be to add a dropdown to allow choosing the grouping criteria of the table, but I'm not yet sure how to tackle this in depth.

This is also a good chance to completely rethink the _Connection details_ popup that is shown when clicking on a specific table entry, since at the moment I feel this is one of the worst looking and most confusing aspects of the app.

PR #1109, opened 2026-04-03
Enhance favorites: support services, programs, and persistence

### LINKED ISSUE #839: Enhance favorites
So far, Sniffnet has been featured by a very **limited** "save as favorite" functionality.

First of all, such a functionality is only applicable to network hosts.
Additionally, favorite network hosts aren't saved across different sessions of the app: this means that every time Sniffnet is restarted or the analysis is interrupted, all the save favorites are lost.

It's finally time to enhance this feature, bringing it at pace with all the other app's advancements of the past months.

Here's a list of requirements for the new, enhanced favorites:
- persist favorites across different runs of Sniffnet
- add an interactive menu showing all the saved favorites, with the possibility to remove each of them individually and add new ones
- allow saving an upper layer service as favorite 
- allow saving a program/process as favorite (depends on #170)
- update notifications to support the new favorites kinds

GyulyVGC:
Enhance favorites:
- persistence across different runs of the app
- support saving upper layer services as favorites
- support saving programs as favorites
- update notifications to support favorite services and programs
- add filters in _Overview_ page tables to only show the saved favorite items, including currently inactive ones (dimmed)

https://github.com/user-attachments/assets/c505f50d-298a-427f-a7f5-da76d20dc1b1

Fixes #839 

***

> [!NOTE]
>
> _The huge amount of files changed is due to the fact that I've also updated all the existing country flags to be more minimal and smaller (circular instead of rectangular)._

***

> [!NOTE]
>
> _The only feature mentioned in #839 that has not been implemented here is having a new page showing per-favorite statistics._ 
> _This is on purpose since I plan to introduce such features for every item regardless its favorite status (see #1096)._


GyulyVGC (commented):
🌟

PR #1152, opened 2026-04-23
Update translations_5.rs

mulecram:
Update DE translations


GyulyVGC:
Hey @mulecram thanks so much for your time, but an update for the German translation was submitted shortly before this PR, and I've merged it today.

Issue #1150, opened 2026-04-22
Traffic chart is broken in GUI


skarlheinz-hue:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### What's the problem?

<img width="1212" height="297" alt="Image" src="https://github.com/user-attachments/assets/351289c5-42b4-478e-a4ff-1f0e989911ed" />

### How did you install the app?

downloaded one of the provided packages

### Operating System

Windows

GyulyVGC:
Hey @skarlheinz-hue thanks for taking the time to open this issue

It seems to be a problem with the graphical renderer in use.

Please try setting the environment variable `ICED_BACKEND=tiny-skia` and retry.
This tells the app to use the fallback graphical renderer.

PowerLos:
The variable was set, and the computer was restarted.
The problem persists.

C:\Users\KHS>set
ALLUSERSPROFILE=C:\ProgramData
APPDATA=C:\Users\KHS\AppData\Roaming
CommonProgramFiles=C:\Program Files\Common Files
CommonProgramFiles(x86)=C:\Program Files (x86)\Common Files
CommonProgramW6432=C:\Program Files\Common Files
ComSpec=C:\WINDOWS\system32\cmd.exe
DriverData=C:\Windows\System32\Drivers\DriverData
HOMEDRIVE=C:
HOMEPATH=\Users\KHS
ICED_BACKEND=tiny-skia

GyulyVGC:
On which version of Windows are you running? Please also verify your graphical drivers are up to date.

PowerLos:
Windows Version is "Microsoft Windows [Version 10.0.19045.6466]" inside VmWare.
Installed VmWare-Tools: 

<img width="192" height="52" alt="Image" src="https://github.com/user-attachments/assets/72b22048-8da2-412c-801a-a098d7e4dff8" />

GyulyVGC:
The fact you're running it inside VmWare could be an explanation of the issue.

skarlheinz-hue:
Ok, i understand, but i have another installation under VmWare with Windows Server 2025 and everything is fine there.

<img width="771" height="274" alt="Image" src="https://github.com/user-attachments/assets/4786d354-dee2-4bea-a7ec-753535e5d6a0" />

GyulyVGC:
I'm afraid I can't do much on my end in these cases, since everything comes down to [iced](https://github.com/iced-rs/iced) rendering capabilities.

[...]

PR #1161, opened 2026-04-27
Traffic chart is broken in GUI #1150

amaa-99:
Workaround for the failed detection of the wgpu backend on Linux systems, leading to falling back to the broken tiny-skia backend (which unable to correctly render the chart areas).

Note that this applies only in the cases where the application is started from its icon (e.g. from the applications/start menu).

Considering that, as per the wgpu documentation, the support is Linux and Android (using Vulkan), the side-effect of adding WGPU_BACKEND=vulkan to the Exec configuration would be, at most, to end up using the broken fallback tiny-skia (in case the vulkan backend is not available). Without that it would always end up using the fallback tiny-skia...

https://github.com/GyulyVGC/sniffnet/issues/1150

### DIFF
--- a/resources/packaging/linux/sniffnet.desktop
+++ b/resources/packaging/linux/sniffnet.desktop
@@ -5,6 +5,6 @@ Name=Sniffnet
 Comment=Application to comfortably monitor your network traffic
 Categories=Network;Utility;
 Icon=sniffnet
-Exec=/usr/bin/sniffnet
+Exec=WGPU_BACKEND=vulkan /usr/bin/sniffnet
 StartupWMClass=sniffnet
 Terminal=false


GyulyVGC:
Hey @amaa-99 thanks for the PR, however I'm not sure about this part

> the side-effect of adding WGPU_BACKEND=vulkan to the Exec configuration would be, at most, to end up using the broken fallback tiny-skia (in case the vulkan backend is not available). Without that it would always end up using the fallback tiny-skia...

As I linked in #1150, there's a similar issue (https://github.com/iced-rs/iced/issues/3138) opened in the iced repo that highlights that there are multiple possible backends for wgpu (the standard renderer).

So it's not just a matter of wgpu vs tiny-skia, but also about wgpu + vulkan vs wgpu + opengl, etc.

If iced itself isn't enforcing choosing vulkan, I don't think we should do so ourselves.
I'm not sure about their selection algorithm, but modifying it in some way feels wrong and may cause more drawbacks than expected.

PR #1181, opened 2026-05-06
feat: support CIDR ranges in imported IP blacklists

TheMasterOfDisasters:
## Summary

This PR adds CIDR range support to imported IP blacklists.

Previously, imported blacklist files only supported exact IPv4/IPv6 addresses. This meant common blacklist sources that publish CIDR ranges, such as Spamhaus DROP-style lists, could not be used directly without manually expanding ranges into individual IP addresses.

With this change, imported blacklists can now include:

- exact IPv4 addresses, for example `8.8.8.8`
- exact IPv6 addresses, for example `2001:db8::1`
- IPv4 CIDR ranges, for example `1.2.3.0/24`
- IPv6 CIDR ranges, for example `2001:db8::/32`
- trailing comments after `;` or `#`

## Motivation

I started using Sniffnet's IP blacklist feature with public blacklist sources and noticed that many of them publish entries as CIDR ranges instead of individual IP addresses.

One example is Spamhaus DROP-style lists, where entries are commonly written like:

```text
1.2.3.0/24 ; SBLxxxx
```

At first I considered expanding those ranges into individual IPs, but that quickly becomes impractical. Larger IPv4 ranges can expand into thousands or millions of addresses, and IPv6 ranges should not be expanded at all.

This change makes the blacklist import more useful for my use case while keeping the behavior general for other users who want to import common blacklist/range-based feeds.

## Implementation

The blacklist now stores:

- exact IP addresses in a `HashSet<IpAddr>`
- CIDR ranges as `IpNet` entries

Lookup checks exact IP matches first, then checks whether the IP belongs to any imported CIDR range.

CIDR ranges are not expanded into individual addresses.

## Testing

Tested with:

```bash
cargo fmt --check
cargo check
cargo test
cargo build --release
```

Also tested locally by importing a blacklist file containing CIDR ranges and verifying that matching traffic is marked as blacklisted.

## Notes

I kept this PR focused on CIDR support for the existing text-based blacklist import. Some public blacklist providers also publish JSON feeds, but I did not include JSON import support here to keep the change small.


GyulyVGC:
Cool! This sounds useful.

Thanks a lot, will review when I have some time.

GyulyVGC (approved):

GyulyVGC:
Merged! I've just done some improvements to performance and parsing. Looks good.

GyulyVGC:
@all-contributors please add @TheMasterOfDisasters for code.

Issue #1042, opened 2026-01-01
Charts should be interactive


GyulyVGC:
Sniffnet has always displayed a line-series chart for incoming/outgoing traffic intensity in the _Overview_ page of the app.
Said chart has various limitations: among the most notable ones, users cannot interact at all with it.
The same limitation also hold for the donut chart reporting agglomerated statistics (introduced in version 1.4.0).

Charts should instead be interactive to convey additional information:
- hovering on each segment of the donut chart should display the relative amount of data
- hovering on each data point of the line-series chart should display the relative amount of data AND the involved items in that specific point in time (network hosts, services, notifications if any, etc.)

***

> [!NOTE] 
>
> _Currently the line-series chart is rendered using `plotters-iced` which is mostly unmaintained and doesn't properly support hover events.
> This means that we'll have to create a custom Iced widget for this feature._

PR #1184, opened 2026-05-10
Experiment/canvas chart prototype

TheMasterOfDisasters:
## Summary

This PR improves chart compatibility with Sniffnet's `tiny-skia` software renderer fallback while preserving the existing default-renderer behavior.

The Overview traffic chart now uses a renderer-compatible Iced geometry widget with hover support. This avoids relying on `plotters-iced2` for the interactive Overview chart path and keeps rendering inside Iced's selected renderer.

For the initial adapter page, the default renderer keeps the existing plotters preview charts. Under `ICED_BACKEND=tiny-skia`, adapter preview charts are replaced with lightweight incoming/outgoing packet-rate indicators. This avoids the tiny-skia geometry/clipping artifacts found when rendering chart geometry inside scrollable adapter cards.

## Why

The original goal was to prototype an interactive chart replacement without making Sniffnet more dependent on GPU-only rendering.

During testing, the Overview chart worked correctly with both the default renderer and `tiny-skia` after switching to an Iced-native geometry widget.

Adapter preview charts exposed a separate tiny-skia issue: chart geometry inside the scrollable adapter list could render, but its clipping/damage tracking did not consistently follow scrolling or page changes. Multiple geometry and Canvas approaches were tested, but the reliable tiny-skia-compatible solution was to avoid chart geometry in that scrollable list.

## Changes

- Added a renderer-compatible Overview line chart widget.
- Added hover detection and tooltip-style value display for Overview traffic points.
- Kept the existing plotters preview charts for the default renderer.
- Added a tiny-skia-specific adapter preview fallback using lightweight text indicators.
- Extended traffic preview data to track incoming and outgoing packet counts separately.
- Removed the experimental adapter Canvas preview chart after validating it was not reliable under tiny-skia in scrollable content.
- Kept packet capture behavior unchanged for normal capture flows.

## Testing

- `cargo fmt`
- `cargo check`
- `cargo test`
- `cargo build --release`
- Manual test with default renderer
- Manual test with `ICED_BACKEND=tiny-skia`


TheMasterOfDisasters:
Hi @GyulyVGC , I opened the PR as a draft for now because it partially addresses #1042 and became broader than expected during implementation.

The initial goal was to prototype an interactive replacement for the Overview line-series chart, since the current `plotters-iced` path is a blocker for hover-based chart interactions. This PR moves the Overview traffic chart to a custom Iced-native widget and adds hover support for per-tick incoming/outgoing values.

While working on it, I ran into renderer compatibility issues with `tiny-skia`, so the PR also includes changes to keep the new chart path compatible with Sniffnet’s software renderer fallback. In particular, the adapter preview charts stay on the existing default-renderer path, while `ICED_BACKEND=tiny-skia` uses lightweight packet-rate indicators to avoid scrollable chart geometry artifacts.

I don’t think this fully closes #1042 yet. It should give us the foundation for future improvements that would close it, especially adding donut segment hover and richer line-chart hover details such as involved hosts, services, and notifications for each point in time.

I’m keeping it in draft until I do more manual testing and add screenshots before requesting review.

GyulyVGC:
Hey @TheMasterOfDisasters thanks for your time implementing this, but I prefer not moving forward with this feature for two reasons:
- iced is replacing `tiny-skia` with `vello_cpu`
- I already have plans to improve charts using [iced_aksel](https://github.com/QuistHQ/iced_aksel) with the help of @mahlquistj

In general, if you are planning to do considerable changes to the codebase, I suggest you to verify that the related GitHub issue isn't assigned to someone yet.

You can also check issues with the https://github.com/GyulyVGC/sniffnet/labels/help%20wanted label to see what I'd love help with.

PR #1197, opened 2026-05-28
Hungarian translations up to 1.5 including yet-to-be-used ones

pml68:
Few small caveats:
1. I'm not *entirely* sure about some of the unused translations due to a lack of context, but I can always come back to those later.
2. I didn't update `subset_characters.txt`. I saw there's a `fonts.sh` script that generates the complete file and does subsetting, but running produced an overall negative diff on the subset_characters file, so I didn't include that. On the other hand, I doubt I should be changing the file manually given the script's existence, so I'll just leave the missing glyphs here: `ő ű`


GyulyVGC:
Thanks so much @pml68 

You need specific tools installed to run the script to update fonts.
Don't worry about it, I'll do it myself.

pml68:
If by "specific tools" you mean `pyftsubset`, I have that covered (though I will note that I have no idea *why* I have it installed in the first place). It's simply that the overall change was -250 lines in the subset file which kinda scared me, and since a lot of it was reordered I didn't really check what actually got *removed*.

pml68:
For future reference, is the `translations` branch the correct one to target for related PRs? I see no mention of it in the Translations issue and past PRs vary.

GyulyVGC:
Usually contributors target main and then I internally use `translations` if I have multiple translations in the merge queue, to open just a grouped-PR to main.
Most of the times it's irrelevant if there aren't merge conflicts.

pml68:
Makes sense. Should I change the target branch to main?

GyulyVGC:
Yeah you can change it to main so we're sure your edits are targeting the most recent commit

pml68:
There we go

GyulyVGC (approved):

GyulyVGC:
@all-contributors please add @pml68 for translation.

PR #1201, opened 2026-06-02
Rename UDP/TCP 2408 service from optimanet to cloudflare-warp

maksimyugai:


GyulyVGC:
As I think you noticed, I already closed similar PRs because the service names Sniffnet infers are from the list maintained by `nmap` and we just rely on it

Issue #944, opened 2025-08-27
Parse TLS `ClientHello` messages to obtain more meaningful domain names


GyulyVGC:
So far Sniffnet has always retrieved domain names simply by performing reverse DNS lookups.
This is problematic because most of the times servers involved in the network connections are behind CDNs that are owned by third party providers, and performing a rDNS will reveal the name of that provider instead the name of the server we're receiving data from.

To partially fix this, we can parse the SNI (Server Name Indication) that is a TLS extension reporting the name of the original server.

When this particular extension isn't available or a connection isn't using TLS, we can fallback to use reverse DNS.

ignoramous:
> also needs to come up with a new strategy to map IP addresses to network hosts

We do this for Rethink DNS + Firewall (an Android app) by running a stub DNS resolver that responds with unique but pseudo IPs (from [RFC6598](https://github.com/celzero/firestack/blob/2191381f/intra/dnsx/alg.go#L1286-L1299) & [RFC8215](https://github.com/celzero/firestack/blob/2191381f/intra/dnsx/alg.go#L1361-L1381)) for A/AAAA/HTTPS/SVCB queries. There's also a knob for users to firewall / block all other IPs (ie, IPs not generated by our stub resolver). It turns out, this was a source of never ending edge-cases and bugs... super, super painful (:

it-lgtm:
I guess [this](https://github.com/OISF/suricata/blob/master/src/detect-tls-sni.c) could be taken as an inspiration to how this parsing could be implemented. As far as I can say, looking at the code, the tls.sni parsing take place as follows:
the matcher matches for tls part, then, when the tls handshake proceeds (when the tls state reaches to client hello state), the related server name is retrieved from the buffer. Quite straightforward.

To help you, I considered contacting the original maintainer of the code, but decided that it isn't necessary(the lines were written possibly ~18 years ago).

Below are some excerpts I gathered by chatting with Gemini:

#### **Q1: From a high level, what is the goal of the `tls.sni` keyword's registration?**

The primary goal is **performance**. Suricata processes enormous amounts of traffic and must be incredibly efficient. The `tls.sni` keyword is registered in a way that supports Suricata's two-phase detection strategy: a very fast initial check (pre-filtering) followed by a more detailed inspection only when necessary. This ensures that the engine doesn't waste time on expensive analysis for rules that clearly won't match.

#### **Q2: How does this two-phase detection work, and why is the `tls.sni` registration split between two functions?**

[...]

PR #1221, opened 2026-06-26
Add a DNS protocol analyzer tab (RFC 1035)

joaopdadv:
## Summary

This PR adds a dedicated **DNS** tab that parses the DNS protocol (RFC 1035)
directly from the bytes of UDP/TCP traffic on port 53 and presents it in real time.

Today Sniffnet only does reverse-DNS lookups to enrich displayed IPs and
identifies services by port number — there is no inspection of the DNS protocol
itself. This fills that application-layer gap.

## What it does

- **Manual DNS parser** over `&[u8]` (no external DNS crate), interpreting:
  - the 12-byte header (ID, QR, OPCODE, AA, TC, RD, RA, RCODE, QD/AN/NS/AR counts);
  - the Question section, with **name compression** support (RFC 1035 §4.1.4);
  - the Answer resource records, with RDATA decoded per type
    (A, AAAA, CNAME, NS, PTR, MX, TXT, SOA).
- **Live log** of queries/responses: time, Q/R, domain, type, RCODE, latency, answers.
- **Resolution latency** by correlating responses to queries via the transaction
  ID + (client, server) endpoint pair.
- **Most-queried domains** ranking.
- **Filters** by record type and by response code (the type filter matches the
  query type *or* any answer record type, e.g. CNAME in a CNAME→A chain).

## Implementation

- New `src/networking/dns/` module (`parser.rs`, `types.rs`).
- DNS messages are extracted from `headers.payload` in `parse_packets.rs` before
  `analyze_headers` consumes the headers, carried to the GUI via a new
  `InfoTraffic::dns_events` field, and folded into `gui::types::dns_state`.
- New `RunningPage::Dns` and `gui/pages/dns_page.rs`; tabs render automatically.

## Testing

- 16 DNS-specific tests (parser, latency correlation, ranking, filters) plus an
  end-to-end test that runs `docs/samples/dns_sample.pcap` through the real
  capture-to-parse pipeline. Full suite: 185 passing, no new warnings.
- Cross-checked against `tcpdump` on the same pcap (identical field interpretation).
- Validated live capturing real DNS traffic.

## Notes / scope

- Encrypted DNS (DoH/DoT) is out of scope — only plaintext DNS on port 53 is parsed.
- The Authority/Additional sections are counted (NSCOUNT/ARCOUNT) but not expanded.


GyulyVGC:
Hey @joaopdadv thanks for your effort but I already have some plans regarding more precise name resolutions as part of #944 

I'll keep this PR open for now and I'll probably take some inspiration, but given the importance of the feature and its impact I'll have to take more time to dig into it and possibly come up with a final implementation myself

PR #1229, opened 2026-07-06
chore: remove dead commented-out code

LeandroPG19:
Removes ~60 lines of commented-out dead code:

- `sample_for_screenshot` in `traffic_chart.rs` (50 lines referencing `/dev/urandom`)
- `get_percentage_string` in `formatted_strings.rs` (12 lines)
- Commented-out IPv6 test cases in `mmdb/country.rs` (4 lines)

These blocks pollute `grep unwrap` audits and give false signals that code paths are live.

```
test result: ok. 171 passed; 0 failed
``


GyulyVGC:
I don't see any value in removing these

PR #1253, opened 2026-07-15
feat(docker): publish multi-architecture images for amd64 and arm64

### LINKED ISSUE #1252: Support linux/arm64 Docker images
## Problem

The published Docker image currently exposes only the `linux/amd64` platform, so ARM64 hosts must use emulation or cannot run a native image.

## Proposed change

Publish a multi-platform OCI manifest containing native `linux/amd64` and `linux/arm64` images.

The implementation uses native GitHub-hosted runners for each architecture, isolates BuildKit caches per architecture, and merges the two platform manifests under the version and `latest` tags.

## Validation

The workflow was tested in a fork. The published image manifest contains both `linux/amd64` and `linux/arm64` variants.

luojiyin1987:
Closes #1252

## Summary

Publish Docker images as a multi-platform OCI manifest for `linux/amd64` and `linux/arm64`.

## Changes

- Build each platform on its native GitHub-hosted runner.
- Push architecture-specific images, then merge them under the version and `latest` tags.
- Isolate GitHub Actions BuildKit caches per architecture.
- Normalize the GHCR image name to lowercase.
- Add a dependency-cache layer and thin LTO to keep ARM64 builds within runner resources.
- Preserve the existing root runtime user so packet capture retains its required capabilities.
- Refuse version-tag overwrites by default; `force_publish` explicitly enables an existing amd64-only version to be migrated.

## Validation

- The workflow completed successfully in this fork.
- `ghcr.io/luojiyin1987/sniffnet:v1.5.0` was inspected and contains `linux/amd64` and `linux/arm64` manifests.
- `docker buildx build --check` completed for `linux/amd64` and `linux/arm64`.
- `git diff --check` passed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,13 +1,9 @@
-# Ignore everything
 *
-
-# Explicitly allow files
-!src/**/*.rs
-!build.rs
 !Cargo.toml
 !Cargo.lock
-!Dockerfile
+!build.rs
 !services.txt
+!src/**/*.rs
 !resources/embedded_icons/**/*.svg
 !resources/DB/*.mmdb
 !resources/fonts/subset/*.ttf
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,13 +3,18 @@ name: Docker
 on:
   workflow_dispatch:
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
 jobs:
-  docker:
-    name: Docker Build
+  check-version:
+    name: Check version
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      packages: write
+    outputs:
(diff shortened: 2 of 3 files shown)


luojiyin1987:
https://github.com/luojiyin1987/sniffnet/actions/runs/29383678261 
image

GyulyVGC:
Hey @luojiyin1987 thanks again for your time but these feel a lot too much changes for a feature that would only need one line change in the arch line of the build file.
As I said in the previous PR you opened, I remember arm64 taking too long without these workarounds.
I'm glad you made it work but I believe the complexities in making it work don't justify the implementation.

luojiyin1987:
Understood, thanks for reviewing it. The native ARM64 runner and dependency caching brought the ARM64 build time close to the amd64 build time, so I’ll keep this implementation in my fork for users who need ARM64 images. No further action is needed here.

PR #1258, opened 2026-07-20
v1.5.1

### LINKED ISSUE #1256: Linux release packages require glibc 2.39, failing on older distros
The prebuilt Linux packages (RPM, DEB, AppImage) fail to start on any distro with glibc < 2.39 — e.g. Fedora 39 (glibc 2.38) reports `/lib64/libc.so.6: version GLIBC_2.39' not found`.

Root cause: the `Package` CI installs `cross` from git `main` (`cargo install cross --git …`), which now uses the Ubuntu 24.04-based `:main` images (glibc 2.39).
Since the `build` job's binary feeds all three Linux packages, they all inherit that floor.
This is a regression from an upstream cross image bump, not a repo change: https://github.com/cross-rs/cross/pull/1779

***

Root cause identified: our CI builds Linux binaries with `cross` installed from git main, whose images moved to an Ubuntu 24.04 base (glibc 2.39) — so the artifacts inherit a glibc 2.39 floor and won't run on older distros like Fedora 39 (glibc 2.38). 

There's no quick pin that fixes it: the only older `cross` image tag is EOL Ubuntu 16.04 (missing needed packages), and the glibc friendly 20.04/22.04 images only ever existed on the moving main tag, so there's nothing stable to pin to.

We also tried `cargo-zigbuild` targeting glibc 2.28: it builds fine after some tweaks, but linking sniffnet (glibc 2.28) against the runner's newer `libasound` / `libpcap` (glibc 2.34+) fails at link time.

### LINKED ISSUE #1257: Reverse DNS lookups are serial (one slow lookup stalls all others)
Reverse DNS resolution runs on a single dedicated thread that processes requests one at a time with the blocking `lookup_addr` (`getnameinfo`).
When an IP has no PTR record or a slow/unreachable DNS server, that call blocks until timeout (~seconds), and every queued lookup waits behind it, causing new network hosts to be temporarily invisible.

GyulyVGC:
Preparing for version 1.5.1 release 🎉

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -58,7 +58,7 @@ jobs:
         run: |
           Write-Host "::group::Npcap SDK"
           $ARCH = "${{ matrix.arch }}"
-          Invoke-WebRequest -Uri "https://npcap.com/dist/npcap-sdk-1.13.zip" -OutFile "$env:TEMP\npcap-sdk.zip" -Verbose
+          Invoke-WebRequest -Uri "https://npcap.com/dist/npcap-sdk-1.16.zip" -OutFile "$env:TEMP\npcap-sdk.zip" -Verbose
           Expand-Archive -LiteralPath "$env:TEMP\npcap-sdk.zip" -DestinationPath "$env:TEMP\npcap-sdk" -Verbose
           $LibPath = switch ($ARCH)
           {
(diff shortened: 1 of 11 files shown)


GyulyVGC (commented):
🚢

GyulyVGC:
@all-contributors please add @jorgemonizribeiro-blip for financial.

PR #1270, opened 2026-08-05
Add IPFIX collector capabilities to Sniffnet

### LINKED ISSUE #303: SniffNet Remote Agent to Monitor the Network Adapter on a Remote Host or Hosts

Setup a Remote Agent for SniffNet, just like Nagios's Remote Agent to Monitor the Network Adapter on a Remote Host or Hosts.
This would be a Option just like in SniffNet when you Pick which Network Adapter you would like to Monitor Network Traffic.
The Remote Agent would be Installed on Windows/Linux and Send Live Data of Network Traffic Information back to SniffNet.

_No response_

GyulyVGC:
Sniffnet can now act as an **IPFIX collector**, receiving and analysing flow records exported by routers, firewalls, and other devices connected to the network.

A new capture source is selectable from the initial page: instead of picking a local network adapter or file, you can bind Sniffnet to a UDP port and let it receive IPFIX flows.
IPFIX flows are mapped onto Sniffnet's existing connection model, allowing you to navigate data from the graphical interface you're already familiar with.

https://github.com/user-attachments/assets/a700622f-e122-4792-ad67-3b216dec953e

Fixes #303


GyulyVGC (commented):
🚢

PR #1275, opened 2026-08-14
fix(inspect): allow favorites for unresolved hosts

cuyua9:
## Summary

- allow favorites-only filtering before reverse DNS resolution completes
- keep country, ASN, and domain filters gated on resolved host data
- add focused regression tests for both behaviors

## Validation

- cargo fmt --all -- --check
- cargo test --bin sniffnet search_parameters

The current early return treats `only_favorites` as a host filter, which hides favorited service/program entries while their host metadata is still unresolved.

### DIFF
--- a/src/report/types/search_parameters.rs
+++ b/src/report/types/search_parameters.rs
@@ -68,10 +68,7 @@ impl SearchParameters {
     }
 
     fn is_some_host_filter_active(&self) -> bool {
-        self.only_favorites
-            || !self.country.is_empty()
-            || !self.as_name.is_empty()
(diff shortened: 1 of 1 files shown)


GyulyVGC (approved):

GyulyVGC:
Hey @cuyua9 good catch, thanks for the fix.

This bug was due to the fact that previously only hosts could be marked as favourites, but this is no longer the case since v1.5.0 introduced favourites support for services and programs too.

GyulyVGC:
@all-contributors please add @cuyua9 for code.

PR #1281, opened 2026-08-23
IPFIX collector: support multiple exporters

GyulyVGC:
> [!NOTE]
>
> _This will be merged to `main` as part of #1270_

Support receiving traffic from multiple IPFIX exporters simultaneously, allowing to filter collected data by exporter in real-time from the _Inspect_ page of the app.

`AddressPortPair` was extended to have an optional `IpfixExporter` field, which carries the exporter identity for IPFIX-based captures and is `None` for standard PCAP-based captures.
This way, the "_same_" 5-tuple network flow can be accounted for as many times as the different exporters observing it.

https://github.com/user-attachments/assets/1cb9e3ae-9f43-4549-b778-70df5d3c7855

### DIFF
--- a/src/gui/pages/connection_details_page.rs
+++ b/src/gui/pages/connection_details_page.rs
@@ -37,7 +37,7 @@ use crate::translations::translations_3::{
     copy_translation, messages_translation, service_translation,
 };
 use crate::translations::translations_5::program_translation;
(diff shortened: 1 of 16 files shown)


GyulyVGC (commented):
🚢

Issue #1037, opened 2025-12-24
Picklist's menu styling issues


GyulyVGC:
Iced's menu styling and rendering for picklist and combobox widgets is currently pretty limited and broken:
- you can't specify a class for the scroller / scrollable
- scrollable style doesn't respond to hover events
- menu borders aren't respected when an option is hovered
- additionally, the combobox widget doesn't let you know the status/number of the currently filtered options, causing Sniffnet to display a longer menu height when filtering a menu with more than 9 total options

https://github.com/user-attachments/assets/30e6110c-351d-4184-ade3-97582083e00e

PR #1284, opened 2026-08-25
Fix oversized combobox menu when options are filtered

jadhavgaurav:
## Summary

Issue #1037 lists five picklist/combobox menu-styling problems. This PR
fixes the one that's actually fixable in this codebase; the other four are
blocked by the vendored `iced` crate's own internals.

## Investigation (iced 0.14.0 / iced_widget 0.14.2)

1. **Can't specify a class for the scroller/scrollable** — not fixable here.
   `overlay::menu::Overlay::new` hardcodes `Scrollable::new(List{...})` with
   no `.class()` call; `menu::Catalog::default_scrollable()` is dead API,
   never invoked anywhere in the crate.
2. **Can't specify scrollbar properties** — same hardcoding; `PickList`/
   `ComboBox` only expose `menu_class`/`menu_style`/`menu_height`, nothing
   for the internal scrollbar.
3. **Scroller style doesn't respond to hover** — scrollbar hover handling
   lives inside `iced_widget::scrollable`, invoked internally by the
   hardcoded `Scrollable` above.
4. **Menu borders not respected on hover** — confirmed real, but the bug is
   in `overlay::menu::List::draw` inside the vendored `iced_widget` crate
   (the hover-highlight quad insets width by `style.border.width` but not
   height/y) — not fixable without patching or upgrading iced.
5. **Combobox doesn't expose filtered count, causing an oversized menu** —
   **fixed**. `combo_box::State::options()` only returns the full unfiltered
   list; sniffnet's own filtering logic was using that total to decide
   whether to cap menu height, so a filtered-down list still got the full
   200px-cap treatment.

## Fix

Added `count_matching_options()` in `src/gui/pages/inspect_page.rs`,
replicating iced's internal `combo_box` matching rule (case-insensitive,
alphanumeric-only substring match per whitespace/punctuation-split query
term), and used it in place of `combo_box_state.options().len()` for the
menu-height check.

## Tests

- 4 new unit tests for `count_matching_options`.
- `cargo build` — success.
- `cargo test` — 180/180 passing (includes the 4 new tests).
- `cargo clippy -- -D warnings` (the command CONTRIBUTING.md/CI use) — clean.
- `cargo fmt --all -- --check` — clean.
- `CHANGELOG.md` updated under `[UNRELEASED]`, noting this is a partial fix
  of #1037 (item 5 only), not the full issue.

## AI-assisted contribution disclosure

This change was produced with AI assistance (Claude Code) and reviewed by
me before submitting, per CONTRIBUTING.md's AI-assistance policy. Items
1-4 are left open in the issue since they need upstream `iced` changes.

## Related issue

Addresses (partially) #1037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All Sniffnet releases with the relative changes are documented in this file.
 
 ## [UNRELEASED]
+- Size the domain/ASN/country/program filter dropdown menu based on the number of currently filtered options rather than the total option count (fixes [#1037](https://github.com/GyulyVGC/sniffnet/issues/1037))
 - Correctly filter by favorites-only before rDNS completes ([#1275](https://github.com/GyulyVGC/sniffnet/pull/1275))
 - Set `Content-Type: application/json` header on remote notifications ([#1266](https://github.com/GyulyVGC/sniffnet/pull/1266))
 
--- a/src/gui/pages/inspect_page.rs
+++ b/src/gui/pages/inspect_page.rs
@@ -446,6 +446,40 @@ fn filter_input<'a>(
         })
 }
 
+/// Counts how many `options` actually match `filter_value`, using the same
+/// matching rule as `iced`'s internal `combo_box` filtering (case-insensitive,
+/// alphanumeric-only comparison, every whitespace/punctuation-separated term
+/// of the query must be contained in the option).
+///
+/// `iced::widget::combo_box::State` only exposes the full, unfiltered list of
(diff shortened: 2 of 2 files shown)


GyulyVGC:
Hey @jadhavgaurav thanks for taking time with the combobox issues.

However, I'd like to fix them all together, and given that the items we cannot fix here (1-4) are the most impactful, I believe that if you want to contribute in this direction it'd be better to directly submit a PR to iced with the fixes for items 1-4, and once those land we can also tackle 5.

PR #1287, opened 2026-08-26
Fix infinite spinning loop when a captured interface goes down

### LINKED ISSUE #1028: Taking down a network interface causes infinite spinning loops
If an existing network interface is taken down, `pcap`'s `next_packet` method will indefinitely throw a generic error `PcapError("The interface disappeared")`, causing our associated loops to start spinning.

This issue manifests in two cases:
- the currently monitored network interface is taken down during an ongoing packet capture
- any of the system network interfaces is taken down while in the initial page of the app (this occurs after #1026)

Even if the interface is later brought back up, the existing capture will still return the error, and a new capture must be re-instantiated to be able to correctly read packets from the interface again.

_Note: this is based on my observation on macOS, but behaviour could be platform-dependent._

jadhavgaurav:
## Summary

`pcap`'s `next_packet()` keeps returning a fatal `PcapError` (e.g. `"The interface disappeared"` on macOS, see also [rust-pcap/pcap#393](https://github.com/rust-pcap/pcap/issues/393)) once the network interface being captured on is taken down. Both packet-capture loops (`parse_packets::packet_stream` and `traffic_preview::packet_stream`) treated any non-`NoMorePackets` error as transient and kept calling `next_packet()` in a tight loop, pegging a CPU core forever instead of stopping.

## Changes

- Added `CaptureType::is_fatal_error`, classifying every `pcap::Error` as fatal except the two expected conditions: `TimeoutExpired` (no traffic in the last read window on a live capture) and `NoMorePackets` (EOF on an offline import).
- Both `packet_stream` threads (device capture and traffic-preview capture) now stop polling as soon as a fatal error is hit, instead of spinning on the same error indefinitely.
- `parse_packets`'s main loop reports the fatal error via a new `BackendTrafficMessage::CaptureError` / `Message::CaptureError`, reusing the existing `CaptureError::Fatal` mechanism already used for capture-start failures, so the failure is surfaced to the user instead of silently spinning.
- `waiting_page` now shows this fatal error even if traffic had already been captured in the session (previously it only showed pre-capture errors, since it bails out early once packets have been seen).

## Notes

- This does not add automatic re-instantiation of the capture once the interface comes back up (the issue mentions this is a separate, not-yet-implemented follow-up); the immediate bug (the busy-loop) is fixed and the failure is now surfaced instead of silently spinning.
- I could only unit-test the error-classification logic (`is_fatal_error`); actually simulating an OS-level interface-down event isn't practical in CI, so that part is verified by manual reasoning about the loop's exit conditions rather than an automated test.
- Disclosure: I used AI assistance (an LLM-based coding agent) to help investigate this issue and draft the fix; I reviewed and tested the change myself before opening this PR.

Fixes #1028

## Test plan
- [x] `cargo build`
- [x] `cargo test` (233 passed)
- [x] `cargo clippy -- -D warnings` (matches CI invocation)
- [x] `cargo fmt --all -- --check`
- [x] Manual review of the loop's exit/continue conditions to confirm the fatal-error path actually breaks the spin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All Sniffnet releases with the relative changes are documented in this file.
 
 ## [UNRELEASED]
 - IPFIX collector capabilities: receive and analyze network traffic from remote devices ([#1270](https://github.com/GyulyVGC/sniffnet/pull/1270) — fixes [#303](https://github.com/GyulyVGC/sniffnet/issues/303))
-- Fix infinite spinning loop when the captured network interface goes down (fixes [#1028](https://github.com/GyulyVGC/sniffnet/issues/1028))
+- Fix infinite spinning loop when the captured network interface goes down ([#1287](https://github.com/GyulyVGC/sniffnet/pull/1287) — fixes [#1028](https://github.com/GyulyVGC/sniffnet/issues/1028))
 - Correctly filter by favorites-only before rDNS completes ([#1275](https://github.com/GyulyVGC/sniffnet/pull/1275))
 - Set `Content-Type: application/json` header on remote notifications ([#1266](https://github.com/GyulyVGC/sniffnet/pull/1266))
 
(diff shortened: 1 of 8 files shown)


GyulyVGC:
It's not reliable to classify all such error classes as fatal errors, and we risk to stop captures that would have otherwise kept working fine.

Until we're not able to isolate the "interface went down" event, there's the risk to abort captures whose errors aren't fatal, which is way worse than the current behavior.

PR #1289, opened 2026-08-27
Open PCAP files from Linux and Windows file managers

### LINKED ISSUE #1288: Allow associating PCAP files to import them with "Open with > Sniffnet"
Sniffnet already allows to import PCAP files in-app, but it'd be awesome to allow users associate such files with Sniffnet to open and analyse them by just clicking on the file icon.

SR0725:
This connects file-manager "Open with" actions to Sniffnet's existing PCAP import flow on Linux and Windows.

Changes:

- accept a PCAP path as a positional CLI argument and start offline analysis
- register `application/vnd.tcpdump.pcap` in the Linux desktop entry
- register `.pcap`, `.pcapng`, and `.cap` in the Windows MSI
- keep the existing `--adapter` startup path and reject conflicting capture sources

macOS is not included in this change. The current app bundle starts through a shell wrapper, while Launch Services delivers an open-document event instead of a file argument. Supporting that path needs a native event bridge rather than another file-association entry.

Verification:

- `cargo test` — 232 passed
- `cargo clippy -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo run -- --help`
- `xmllint --noout resources/packaging/windows/setup.wxs`

I could not build the Windows MSI locally from macOS, so the Windows packaging job remains the platform-specific validation for that part.

AI assistance: I used an AI coding assistant while investigating and implementing this change. I reviewed the diff and can explain each change.

Related to #1288.


GyulyVGC:
Hey @SR0725 thank you!

Nice to see that implementing support for this is quite simple.
Ideally I'd like to implement support for all the first-class supported platforms (including macOS) all at once, but this can be a good starting point.

PR #1290, opened 2026-08-28
Various improvements and fixes

GyulyVGC:
Various improvements and fixes:
- Show output file path in Overview page when exporting a PCAP file
- Fix the app freezing and exhausting memory when importing a PCAP file containing a large time gap between consecutive packets
- Restrict the PCAP export file to have a valid PCAP extension, so that files of other types can't be overwritten
- IPFIX collector: don't parse zero as a valid port number, and limit number of cached templates and baselines


GyulyVGC (commented):
🚢

PR #1291, opened 2026-08-28
Reduce font size in tooltips

GyulyVGC:
Minor improvement: decrease font size in all the app tooltips to reduce visual clutter


GyulyVGC (commented):
🚢

PR #1296, opened 2026-08-31
Allow opening PCAP files with Sniffnet from the OS

### LINKED ISSUE #1288: Allow associating PCAP files to import them with "Open with > Sniffnet"

m-altaifi:
## Summary
- adds a positional PCAP file argument for file-open launches
- registers PCAP file associations in Linux, macOS, and Windows packaging
- forwards macOS file-open arguments through the privilege wrapper
- keeps adapter startup explicit by making `PCAP_FILE` conflict with `--adapter`

## Testing
- [x] cargo fmt --check
- [x] cargo check
- [x] git diff --check
- [x] parsed `resources/packaging/macos/Info.plist` as XML
- [x] parsed `resources/packaging/windows/setup.wxs` as XML
- [x] cargo test cli::tests - blocked locally by missing `wpcap.lib`
- [x] verified opening a `.pcap` file from the OS launches Sniffnet
- [x] verified regular in-app PCAP import still works

Fixes #1288


GyulyVGC:
Thanks!
I'll look into this when I have some time.

Issue #97, opened 2023-03-04
Suggest me keyboard shortcuts to introduce


GyulyVGC:
I started introducing keyboard shortcuts to improve the efficiency of use and the overall user experience.

I have a few ideas about where to apply shortcuts but I'm unsure of the shortcut keys to be used, so feel free to give me some suggestions.

- Quit the application: ctrl+Q
- Start the sniffing process from the initial page: enter
- Close settings popup: esc
- Switch from a page to another (from overview to notifications and between different settings pages): tab (or maybe ctrl+tab?)

More shortcuts will be added and you can even suggest your own.

Rianico:
For `Open settings`, how about `ctrl+,` (`cmd+,` on macOS)?
On mac, almost all apps open their settings using `cmd + ,`.
On windows, I don't know so much, but [windows terminal](https://allthings.how/how-to-use-windows-terminal-keyboard-shortcuts/) use analogous `ctrl + ,` to open "Open Windows Terminal Settings UI".

GyulyVGC:
Hi @Rianico. 
Thanks for your advice. 
I think I'll implement your suggestion, since ctrl+S was kinda misleading because it's usually used to save something. Thank you!

Rianico:
> Hi @Rianico. Thanks for your advice. I think I'll implement your suggestion, since ctrl+S was kinda misleading because it's usually used to save something. Thank you!

  You're welcome. I would also like to thank you for this powerful project.

corneliusroemer:
Made a suggestion in a self-standing issue to simplify discussion of this suggestion: 
> Bind the `cmd/ctrl` + `-` and `cmd/ctrl` + `-` shorcuts to increase decrease the zoom levels

- #554

corneliusroemer:
- [ ] Add shortcut to show keyboard shortcuts, the canonical one is `cmd/ctrl` + `?` ~= (on US keyboard) `cmd/ctrl` + `shift` + `/`
- [x] Add shortcut to pause recording, once #551 is implemented, suggested shortcut `cmd+enter` or `cmd+F` (capital F for freeze)
- [ ] Switching tabs "Overview", "Inspect", "Notifications" with shortcuts `CMD+1`/`CMD+2`/`CMD+3`

PR #1298, opened 2026-08-31
feat: add running page keyboard shortcuts

m-altaifi:
## Summary
- adds Ctrl+1, Ctrl+2, and Ctrl+3 shortcuts for the Overview, Inspect, and Notifications pages
- keeps the shortcuts scoped to the running view so settings and modals are not changed behind overlays
- adds unit coverage for direct page switching and unread notification reset behavior

## Testing
- cargo fmt --check
- cargo check
- cargo test test_correctly_switch_running_page_with_shortcut (blocked locally: `link.exe` cannot find `wpcap.lib`)

Refs #97

### DIFF
--- a/src/gui/sniffer.rs
+++ b/src/gui/sniffer.rs
@@ -229,13 +229,13 @@ impl Sniffer {
                         Key::Named(Named::ArrowLeft) => Some(Message::ArrowPressed(false)),
                         Key::Named(Named::ArrowRight) => Some(Message::ArrowPressed(true)),
                         Key::Character("1") => {
-                            Some(Message::RunningPageShortcut(RunningPage::Overview))
+                            Some(Message::ChangeRunningPage(RunningPage::Overview))
                         }
                         Key::Character("2") => {
-                            Some(Message::RunningPageShortcut(RunningPage::Inspect))
+                            Some(Message::ChangeRunningPage(RunningPage::Inspect))
                         }
                         Key::Character("3") => {
-                            Some(Message::RunningPageShortcut(RunningPage::Notifications))
+                            Some(Message::ChangeRunningPage(RunningPage::Notifications))
                         }
                         Key::Character("-") => Some(Message::ScaleFactorShortcut(false)),
                         Key::Character("+") => Some(Message::ScaleFactorShortcut(true)),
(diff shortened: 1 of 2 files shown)


src/gui/types/message.rs
GyulyVGC, line 78: We already have the `ChangeRunningPage` message. Why not reusing it?

src/gui/sniffer.rs
```diff
@@ -1300,6 +1310,12 @@ impl Sniffer {
         }
     }
 
+    fn running_page_shortcut(&mut self, running_page: RunningPage) {
```
GyulyVGC: I think we can update `change_running_page` with these side-effects guard?

GyulyVGC (changes requested):

src/gui/types/message.rs
m-altaifi: Good call — done in 7e89d0a. Dropped `RunningPageShortcut` entirely; Ctrl+1/2/3 now emit `ChangeRunningPage` directly.

src/gui/sniffer.rs
m-altaifi: Done — the guard is now an early return in `change_running_page`, so `running_page_shortcut` is gone and the tab-bar click path gets the same protection.

One thing to flag: `test_conf` and `test_toggle_thumbnail` were driving `ChangeRunningPage` from states the guard now rejects (no running page, or settings open), so I added explicit setup to both. Those states are unreachable from the real UI, but shout if you'd rather the guard were narrower.

m-altaifi:
The requested refactor is now in the latest commit: the shortcuts reuse \ChangeRunningPage\, and the shared handler owns the side-effect guard. The related tests now set the state required by that guard. Could you please re-review when convenient?

GyulyVGC:
Hey, thanks for the reminder

I already reviewed actually and everything looks good, but it'll take some time to merge this since I'm working on other items right now.

Hopefully I'll ship this with the next version of Sniffnet

GyulyVGC:
Hey @m-altaifi I think this should be extended to work also with setting pages as I'm doing for the `tab`-based keyboard shortcut

Issue #1269, opened 2026-08-04
Support IGMP


GyulyVGC:
Support IGMP protocol traffic and message types

It should now be possible to support IGMP since it was added in [etherparse 0.21](https://github.com/JulianSchmid/etherparse/releases/tag/v0.21.0)

Part of #1069

Issue #1070, opened 2026-02-14
Parse VLAN headers


GyulyVGC:
Despite we already correctly include packets with a VLAN header, we are not parsing it.

This issue wants to track the need to parse VLAN headers and show information about VLAN tags.

Issue #1038, opened 2025-12-24
Extensive integration testing


GyulyVGC:
This issue wants to track the need for extensive integration testing of the whole application.

Some unit tests already assert the behaviour of the app in its most critical aspects, but right now there's nothing in place even remotely close to appropriate and thorough integration tests.
The scarcity of testing was especially due to two aspects:
1. the challenge to have a reproducible and testable stream of network traffic
2. the lack of a framework to properly test Iced-based GUIs

Now, both of these limitations seem to be gone.
Point 1. is not a problem anymore because Sniffnet now supports PCAP files import (#795) — this means that we can test packets parsing functionalities against a capture file crafted for the purpose.
Point 2. is also obsolete, since starting from [Iced 0.14](https://github.com/iced-rs/iced/releases/tag/0.14.0) they introduced various testing facilities that are game changers in this regard:
- headless mode testing (https://github.com/iced-rs/iced/pull/2698)
- time travel debugging (https://github.com/iced-rs/iced/pull/2910)
- first-class end-to-end testing: test recorder, runtime emulator, program presets, Ice test syntax, and selector API (https://github.com/iced-rs/iced/pull/3059)

Issue #489, opened 2024-03-30
Benchmark Sniffnet performance


GyulyVGC:
This issue wants to track the need for a benchmark to accurately measure the speed of the app, measured in packets per second (pps) that can be handled without dropping frames.

This is necessary to find out ways that can help improving the performance even further.

> Today I tried testing Sniffnet with `iperf` on a different machine and I could observe that the current packets per second limit is higher than 100k:
>
> ![Screenshot 2024-01-09 at 12 26 24](https://github.com/GyulyVGC/sniffnet/assets/100347457/d1c49afe-a65d-433b-a16b-522c118fc8b3)
>
> _Originally discussed in #373_

GyulyVGC:
Since the introduction of PCAP files import in version 1.4, we now have a more reliable and stable way of assessing the app performance without relying on Internet connectivity and network adapter speed limits.
It's in fact enough to import a large PCAP file and divide the total number of parsed packets by the time it takes to parse the file to obtain an accurate packet per second rate.

An example of this is shown in [this blog post](https://sniffnet.net/news/v1.4/).

***

**_Update_**: and this is a video showing the same file of the blog post parsed by another computer of mine running a Debian distribution on a VM.

https://github.com/user-attachments/assets/6fcb62fd-c68c-42cd-a7db-ea99af522e52

This mini PC is way faster than the PC I used for the blog post video.

<details>

<summary>PC & VM details</summary>

Beelink SER7 Mini-PC, Ryzen 7 7840HS (8C/16T, up to 5.1GHz) Mini Computer, AMD Radeon 780M, 32GB DDR5+1TB SSD Micro Desktop, Mini Gaming PC 4K Quad Displays HDMI+DP+Dual USB4, 2.5Gbps LAN, WiFi 6

<img width="625" height="232" alt="Image" src="https://github.com/user-attachments/assets/d0ccb880-a2a5-4ef0-9585-075455e2d9f5" />

</details>

As you can see it processes 2.6 million packets in just about 6 seconds, reaching a mad rate of **more than 400k packets per second**!

GyulyVGC:
Update: performance remains pretty much the same (around about 6 seconds to parse that file) even after #957.

GyulyVGC:
Update: no significant difference in performance even after #992.

shishir-cyber:
Hi @GyulyVGC, I'd like to work on this. I'm planning to add criterion as a dev-dependency and create a benches/parsing.rs benchmark.
[...]

PR #1299, opened 2026-08-31
Extract packet parsing logic in a new library: `sniffnet-packet-parser`

GyulyVGC:
Extract packet parsing logic in a new library (`sniffnet-packet-parser`).

This is made possible by restructuring the project skeleton making it a Rust workspace instead of a standalone package.
The main package is still the Sniffnet app binary, but now a new `lib` folder is also present to host libraries including shared functionalities.

For now, just the `sniffnet-packet-parser` library is in there, and I judged it worth it to make this change since now the packet parsing logic is shared both by Sniffnet and [`sniffnet-agent`](https://github.com/GyulyVGC/sniffnet-agent).
Having such logic centralised will allow to introduce support for new packet parsing capabilities (e.g., #1070, #1269) with a single source of truth.

Extracting relevant logic to new libraries is also what will open the way for benchmarking and integration testing (#489, #1038).


GyulyVGC (commented):
🚢

PR #1302, opened 2026-09-03
Show VLAN ID of connections (IEEE 802.1Q)

### LINKED ISSUE #1070: Parse VLAN headers

GyulyVGC:
Parse VLAN headers and show VLAN ID for tagged network connections.
This PR covers both the standard PCAP-parsing path and the IPFIX collector.

Fixes #1070


GyulyVGC (commented):
🚢

PR #1304, opened 2026-09-05
IPFIX collector: support ARP flows

GyulyVGC:
Sniffnet already supports ARP traffic in the standard PCAP-parsing path.
This PR adds support for ARP even for the IPFIX collector based on the exported Ethernet type of the flows.


GyulyVGC (commented):
:ship:

PR #1306, opened 2026-09-07
Show info about packets dropped by the network adapter

### LINKED ISSUE #1079: Show number of packets dropped by the network interface
Currently we're only showing packets that are dropped by Sniffnet, but no info about packets dropped at the network interface level.

See [pcap::Stat](https://docs.rs/pcap/latest/pcap/struct.Stat.html) for more info.

Thanks @farsheed for making me realise this.

GyulyVGC:
Sniffnet already reported _"dropped packets"_ without making any distinction between PCAP-dropped and adapter-dropped.

The app only reported PCAP-dropped packets, while now _"dropped packets"_ are the sum of both, and in case of a value greater than 0, a tooltip will appear on hover to show both amounts.

Fixes #1079 

dropped_info


GyulyVGC (commented):
:ship:

PR #1310, opened 2026-09-13
Expanded view

### LINKED ISSUE #884: Allow reducing / hiding the app header and footer
In order to take less unneeded space and let users focus on what really matters, Sniffnet should provide settings to allow reducing and / or hiding the app header and footer.

GyulyVGC:
Add an option in general settings to activate a new **_expanded view_**, where the app header is slimmer and the footer is completely hidden, allowing users to focus on what matters most.

Fixes #884 

https://github.com/user-attachments/assets/d0ee7dbb-9d41-4d58-bb30-cbd95f5231df


GyulyVGC (commented):
🚢

PR #1311, opened 2026-09-15
Add Ubuntu PPA source packaging and publication workflow

GyulyVGC:
Adds Debian source packaging for Ubuntu 26.04 amd64 and a manually triggered GitHub Actions workflow to sign and upload releases to the Sniffnet PPA.

Uses tagged releases with locked, vendored dependencies.


GyulyVGC (commented):
🐧

